PR #180, opened 2016-06-21
setting initial viewpoints for export tiles and generate gdb samples

ldanzinger:


michael-tims:
Looks good.

PR #193, opened 2016-06-28
Kous3106/symbols sample

khajra:
@JamesMBallard Please review.

1. Added symbols samples for qml and c++
2. Updated the extrusion samples for c++ and qml to add some space between the graphics and also color them differently


JamesMBallard:
Looks good. Very nice!

PR #198, opened 2016-06-29
c++ and qml samples for distance composite scene symbol

khajra:
Added both samples.


khajra:
@JamesMBallard Please review.

JamesMBallard:
:+1:

PR #234, opened 2016-07-27
Nand6028/loadable service info changes

nandinirao:
Assigned to @michael-tims 

Sample changes for non-loadable service info objects and loadable tasks.


michael-tims:
Looks good.

PR #249, opened 2016-08-15
updating location display

ldanzinger:
@michael-tims please review and merge. Cleaning up the way the sample works so it is consistent with other Runtime SDK teams


michael-tims:
Looks good.

PR #252, opened 2016-08-24
updating to > 2.0 web map

ldanzinger:
@michael-tims please merge. The old ID we had was for an old web map. The new ID is > 2.0


michael-tims:
Looks good.

PR #253, opened 2016-08-26
Use the STATE_ABBR field.

JamesMBallard:
Assigned to @khajra 

Please review. We can use STATE_ABBR field now.

cc: @ryankim86


khajra:
:+1:

PR #257, opened 2016-08-30
Mobile map (search and route) sample

ryankim86:
Adding the Mobile Map (search and route) sample. @khajra could you please review this? Thank you!


ArcGISRuntimeSDKQt_CppSamples/Maps/MobileMap_SearchAndRoute/MobileMap_SearchAndRoute.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+void MobileMap_SearchAndRoute::componentComplete()
+{
+    QQuickItem::componentComplete();
+
+    // find QML MapView component
+    m_mapView = findChild<MapQuickView*>("mapView");
+    m_mapView->setWrapAroundMode(WrapAroundMode::Disabled);
+
+    m_fileInfoList = QDir(QDir::homePath() + "/ArcGIS/Runtime/Data/mmpk/").entryInfoList();
```
khajra: This might cause some problems on ios.  It is better to use paths from qml and expose them as properties and then using them from c++.  Some of the samples have that mechanism.  You can look at those for examples.

ArcGISRuntimeSDKQt_CppSamples/Maps/MobileMap_SearchAndRoute/MobileMap_SearchAndRoute.cpp
```diff
@@ -0,0 +1,366 @@
...
+{
+    if (index < m_fileInfoList.length())
+    {
+        // check if file is a .mmpk file
+        if (m_fileInfoList[index].completeSuffix() == "mmpk")
+        {
+            // create a new MobileMapPackage
+            MobileMapPackage* mobileMapPackage = new MobileMapPackage(m_fileInfoList[index].absoluteFilePath());
+            // load the new MMPK
+            mobileMapPackage->load();
```
khajra: It's better if this call is made after the connection is setup.  So, you can move it after the connection logic.  There might be a chance that the call completes before it hits the connection logic if it comes later.  However, if it setup before the call, then it is safer.

khajra:
@ryankim86 :+1:

PR #270, opened 2016-09-12
The task id is now included.

JamesMBallard:
Assign to @nandinirao 

Use the updated signal which passes back the task id on the signal now.

Merging to v.next.


nandinirao:
:+1:

PR #274, opened 2016-09-12
This property is now called 'ranges'.

JamesMBallard:

This is now called "ranges()".


michael-tims:
@JamesMBallard The same changes are needed for the QML sample.

JamesMBallard:
@michael-tims it looks like they're all added via a default property in the Qml sample, the property name is not used.

https://github.com/Esri/arcgis-runtime-samples-qt/blob/james/Ranges/ArcGISRuntimeSDKQt_QMLSamples/Scenes/DistanceCompositeSymbol/DistanceCompositeSymbol.qml#L93-L138

michael-tims:
@JamesMBallard Oops, I was looking at master, not v.next.  All good.

PR #279, opened 2016-09-18
Set accessoryButtonHidden to true. Also set the CalloutData image thr…

nandinirao:
…ough the QImage setter

Assigned to @michael-tims


michael-tims:
Looks good

PR #281, opened 2016-09-19
turn off attribution text in the sceneView (shown in map)

lsmallwood:
@ldanzinger - please review and merge


ldanzinger:
nice, that works well 👍

PR #289, opened 2016-09-24
Michael tims/add signal connect context

michael-tims:
Assign to @ldanzinger 

Provide context reference for lambda connections.  This helps to disconnect connections automatically with a particular sample is destroyed.


ldanzinger:
👍

PR #292, opened 2016-09-28
Michael tims/generate export tile crash on destroy

michael-tims:
Assign to @ldanzinger 

This fixes a crash for the Sync Features and Export Tiles samples where one could attempt to switch to a different sample while a job is in progress.  We must explicitly disconnect from the job's signals in this case, because the QML engine hasn't fully destructed the Job type (seems like a Qt bug).


ldanzinger:
looks good

PR #294, opened 2016-10-01
Sample updates for IdInfo

michael-tims:
Assign to @JamesMBallard 

Updating samples to use IdInfo.  Also fixed up crashing that can occur when switching samples during generate.


JamesMBallard:
:+1:

PR #303, opened 2016-10-06
Add returnPopupsOnly

michael-tims:
Assign to @nandinirao 

Removing `IdentifyReturns` enum and replacing with `returnPopupsOnly`


nandinirao:
Looks good.

PR #310, opened 2016-10-11
removing workaround for reverse geocode label issue

ldanzinger:
@michael-tims please review and merge. The underlying data had an issue, so that has been updated, and the workaround should be removed from the sample code. The label should now come back fully populated


michael-tims:
Looks good.  Much cleaner.

PR #314, opened 2016-10-13
Don't use static SR method for now (iOS issue)

lsmallwood:
@jtrieu please review and merge the fix for running on iOS


lsmallwood:
Don't need this anymore. Closing

PR #320, opened 2016-10-20
Provide OpenGL context version for 3D on Linux

michael-tims:
Assign to @JamesMBallard 

Specify OpenGL context version for 3D on Linux.


JamesMBallard:
:+1:

PR #323, opened 2016-10-24
Update tests to drop QSharedPointer usages.

JamesMBallard:
Assign to @michael-tims 

Update the samples for the new API changes. We no longer return a QSharedPointer which will be easier for users to use.

I ran all updated samples and found no issues.

`OpenExistingMap` in the widgets test app was out of date. I fixed it too.

Merging to v.next.


michael-tims:
Looks good.

PR #326, opened 2016-10-27
Call the setter for the scene properties on the renderer

nandinirao:
Assigned to @michael-tims 

Calling the setter for scene properties on renderer.


michael-tims:
Looks good.

PR #336, opened 2016-10-31
Cleanup MissionData and move class into Animate3DSymbols files

michael-tims:
Assign to @JamesMBallard 

Moving the MissionData class to live inside the Animation3DSymbols files.  this was the code for the class can be displayed in the sample.


JamesMBallard:
:+1:

PR #338, opened 2016-11-02
Update the minimum supported version of macOS to 10.11.

JamesMBallard:
Assign to @michael-tims 

Update `QMAKE_MACOSX_DEPLOYMENT_TARGET` to 10.11 and add this to the widget samples as well.


michael-tims:
Looks good.

PR #339, opened 2016-11-02
The data was updated so it's always "AddressLabel" now.

JamesMBallard:
Assign to @nandinirao.

This fixes the problem when geocoding with the routing SF map. The data was recently changed so the label is always `AddressLabel`. I refactored the logic a bit too. The Qml sample does not have this problem.


nandinirao:
Ok, looks good.

PR #341, opened 2016-11-03
James/set viewpoint changes

JamesMBallard:

This changes all the 3D samples to use an initial viewpoint rather than zooming in on launch.


michael-tims:
Looks good.

PR #344, opened 2016-11-04
Fixed various sample issues

jtrieu:
@lsmallwood please review and merge


ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/Simple_Renderer/Simple_Renderer.cpp
```diff
@@ -76,7 +76,7 @@ void Simple_Renderer::componentComplete()
     m_mapView->setMap(m_map);
 
     // set viewpoint using the two farthest points as an envelope with padding
```
lsmallwood: (this would be optional for readability) could change comment to "with 50m padding" or maybe create a local var called "paddingM = 50;"?

PR #346, opened 2016-11-07
James/scale padding

JamesMBallard:

This wasn't working on iPhone. It's because as the screen dimensions become more rectangular than square (either way) the padding is amplified. This is a quick a dirty formula to scale the padding value so the "Geometry and Padding" option will always work no matter the screen dimensions.


michael-tims:
The code seems ok, but I'm not sure whether the sample is working "as designed" or not.  And maybe we should just leave as is.

JamesMBallard:
That's fine. Let's close this for now and I will confirm.

michael-tims:
👍

PR #349, opened 2016-11-08
Set the viewpoint once map loads

nandinirao:
Assigned to @michael-tims 

Setting the viewpoint after map loads with a scale. Please merge.


michael-tims:
Fixed.  Merging...

PR #350, opened 2016-11-09
Use sync version of setViewpointCamera

michael-tims:
Assign to @JamesMBallard 

Call the synchronous method setViewpointCameraAndWait for 3D samples when appropriate.  Changed up the "Animate 3D Symbols" sample to use the sync method instead of the async method.


JamesMBallard:
Looks good. Thanks @michael-tims

PR #351, opened 2016-11-09
Add depth so the cone doesn't appear flat and remove the pitch so it …

JamesMBallard:
…appears right side up.

Assign to @khajra. Please review. The cone appears correctly with this change.


khajra:
@JamesMBallard 👍

PR #358, opened 2016-11-11
fixing plist names

ldanzinger:
@JamesMBallard please merge. The names in the plist were incorrect, causing me to not be able to deploy to iOS


JamesMBallard:
Good find. :+1:

PR #371, opened 2016-12-02
Change name to match with cpp sample.

khajra:
Change name.


khajra:
@JamesMBallard Please review.

PR #373, opened 2016-12-05
adding clip into sample so the callout is clipped by the map view

ldanzinger:
@nandinirao please review and merge


nandinirao:
Looks good,

PR #374, opened 2016-12-05
adding clip so that the callout does not display beyond the map view

ldanzinger:
@nandinirao please review and merge


nandinirao:
Looks good.

PR #377, opened 2016-12-07
added proper margins.

khajra:
Fixed the margins so that the controls don't cover the attribution text.


khajra:
@tdunn Can you please review?

PR #378, opened 2016-12-07
updated margins.

khajra:
Fixed the margins so that the controls don't overlap the attribution text.


khajra:
@tdunn Can you please review?

PR #384, opened 2016-12-15
Change sample to use web map Url instead of PortalItem

lsmallwood:
@ldanzinger please review the change to the sample


ldanzinger:
@lsmallwood waiting to here from Thomas about the doc snippet before merging

lsmallwood:
@ldanzinger - having spoken to @tdunn we should just merge and then check the guide doc

ldanzinger:
looks good

PR #389, opened 2016-12-19
updating viewpoint to use center/scale

ldanzinger:
@lsmallwood can you please merge? Switching from extent to center/scale viewpoint to better accommodate differing form factors

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/DisplayGrid/DisplayGrid.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/DisplayGrid/DisplayGrid.cpp
@@ -25,6 +25,8 @@
 #include "MGRSGrid.h"
 #include "SimpleLineSymbol.h"
 #include "TextSymbol.h"
+#include "Viewpoint.h"
+#include "Point.h"
 
 using namespace Esri::ArcGISRuntime;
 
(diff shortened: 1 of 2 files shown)


ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/DisplayGrid/DisplayGrid.cpp
```diff
@@ -67,7 +69,7 @@ void DisplayGrid::componentComplete()
 
   // Create a map using the imagery basemap
   m_map = new Map(Basemap::imagery(this), this);
-  m_map->setInitialViewpoint(Viewpoint(Envelope(-11018849.838073004, 4906181.949903341, -9653433.562293354, 5930244.156738078, SpatialReference(102100))));
+  m_map->setInitialViewpoint(Viewpoint(Point(-10336141.70018318, 5418213.05332071, SpatialReference(3857)), 6450785));
```
lsmallwood: I think it is would be clearer for the user if you assign `6450785` to a variable 1st. E.g.

`int targetScale = 6450785` or whatever

similarly, I'd say it is nice to create a var with the SpatialRef named in some way rather than referring by code which the user has to go and look up. E.g.:

`SpatialReference webMercator = SpatialReference(3857)` or whatever it happens to be...

lsmallwood:
just the 1 comment

PR #395, opened 2016-12-20
removing android files

ldanzinger:
@JamesMBallard please review and merge. I noticed that these android files are incorrectly included in  a few samples. It is only in the master branch


JamesMBallard:
Good catch. Thanks for cleaning it up.

PR #397, opened 2016-12-30
Add a function to format the state names entered by the user so it be…

JamesMBallard:
…tter matches what the service expects.

@lsmallwood please review the changes. This is merging to v.next.


lsmallwood:
Looks good  merging

PR #398, opened 2017-01-04
Do not recreate graphics when mission changes

michael-tims:
Assign to @ldanzinger 

Updated sample so it doesn't recreate graphics each time the mission changes.  Instead it just updates the graphics' geometries and attributes.  Also cleaned up some code a bit in the process and got the 2D symbol on the overview map working.


ldanzinger:
looks good 👍 

nice cleanup

PR #400, opened 2017-01-05
Fixing so that we aren't setting on a copy

ldanzinger:


michael-tims:
Thanks.  This is good.

PR #407, opened 2017-01-13
We scale this in the Api now.

JamesMBallard:
Assigned to @michael-tims. Merging to v.next. Please review.

We now calculate this correctly in the Api.

cc: @lsmallwood


michael-tims:
Looks good.

PR #415, opened 2017-01-18
enhancing calendar to increment dates if you pick a from date that is…

ldanzinger:
enhancing calendar to increment dates if you pick a from date that is newer than the current to date, or a to date that is older than the current from date

@JamesMBallard please review and merge


JamesMBallard:
:+1:

PR #419, opened 2017-01-23
adding new import paths into the main.cpp

ldanzinger:
@michael-tims can you please review and merge? This pulls in the plugins using the new SDK structure. Seeing that you are using our PRF/PRI (which the samples are), these will all be defined


michael-tims:
Looks good.

PR #424, opened 2017-01-26
Updating import path syntax 

ldanzinger:
@michael-tims can you please review and merge? These changes are needed in case there are spaces in the location of the QML plugin


michael-tims:
Looks good.

PR #430, opened 2017-02-08
The additional message is more useful in these cases.

JamesMBallard:
Use a more user-friendly error message in case the job fails.


JamesMBallard:
Assigned to @ldanzinger  Please review.

PR #433, opened 2017-02-15
Lsmallwood/qml search symbols

lsmallwood:
@ldanzinger please review the QML version of the sample


ldanzinger:
looks good, nice work 👍

PR #434, opened 2017-02-16
setting initial viewpoint before load cycle begins

ldanzinger:
@JamesMBallard can you please merge? No issue, just found this bug where the initial viewpoint was being set after the load cycle begins, causing the initial viewpoint to not work.


JamesMBallard:
:+1:

PR #443, opened 2017-03-16
Kous3106/ls

khajra:
Adding the local server services sample.


khajra:

PR #465, opened 2017-04-03
start the service directly if server running already.

khajra:
Add a call to start the service directly if the server is running already.


khajra:

PR #500, opened 2017-04-25
Ldanzinger/vectorlayer updates

ldanzinger:
@lsmallwood please review and merge. Updating URLs and using static factory methods/subclasses where it makes sense.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Layers/VectorTiledLayerUrl/VectorTiledLayerUrl.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Layers/VectorTiledLayerUrl/VectorTiledLayerUrl.cpp
@@ -51,7 +51,7 @@ void VectorTiledLayerUrl::componentComplete()
 
   //! [display vector tiled layer]
   // create a vector tiled basemap
-  ArcGISVectorTiledLayer* vectorTiledLayer = new ArcGISVectorTiledLayer(QUrl("http://www.arcgis.com/home/item.html?id=dcbbba0edf094eaa81af19298b9c6247"));
+  ArcGISVectorTiledLayer* vectorTiledLayer = new ArcGISVectorTiledLayer(QUrl("http://www.arcgis.com/home/item.html?id=e19e9330bf08490ca8353d76b5e2e658"));
   Basemap* basemap = new Basemap(vectorTiledLayer, this);
   // create a new map instance
   m_map = new Map(basemap, this);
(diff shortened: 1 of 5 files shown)


ArcGISRuntimeSDKQt_CppSamples/Layers/VectorTiledLayerUrl/VectorTiledLayerUrl.cpp
```diff
@@ -70,13 +70,13 @@ void VectorTiledLayerUrl::changeBasemap(QString basemap)
...
+      vectorTiledLayer = new ArcGISVectorTiledLayer(QUrl("http://www.arcgis.com/home/item.html?id=e19e9330bf08490ca8353d76b5e2e658"));
     else if (basemap == "Streets")
-      vectorTiledLayer = new ArcGISVectorTiledLayer(QUrl("http://www.arcgis.com/home/item.html?id=4e1133c28ac04cca97693cf336cd49ad"));
+      vectorTiledLayer = new ArcGISVectorTiledLayer(QUrl("http://www.arcgis.com/home/item.html?id=a60a37a27cc140ddad15f919cd5a69f2"));
     else if (basemap == "Night")
-      vectorTiledLayer = new ArcGISVectorTiledLayer(QUrl("http://www.arcgis.com/home/item.html?id=bf79e422e9454565ae0cbe9553cf6471"));
+      vectorTiledLayer = new ArcGISVectorTiledLayer(QUrl("http://www.arcgis.com/home/item.html?id=92c551c9f07b4147846aae273e822714"));
     else if (basemap == "Dark Gray")
-      vectorTiledLayer = new ArcGISVectorTiledLayer(QUrl("http://www.arcgis.com/home/item.html?id=850db44b9eb845d3bd42b19e8aa7a024"));
+      vectorTiledLayer = new ArcGISVectorTiledLayer(QUrl("http://www.arcgis.com/home/item.html?id=5ad3948260a147a993ef4865e3fad476"));
```
lsmallwood: we can potentially end up with a nullptr for the vectorTiledLayer here (because there is no else condition). I don't know if that could crash? Might be worth picking one of the options as a default that we fall back to if the received string doesn't match.

The QString in the argument should also be const &

ArcGISRuntimeSDKQt_QMLSamples/Layers/VectorTiledLayerUrl/VectorTiledLayerUrl.qml
```diff
@@ -67,16 +67,16 @@ Rectangle {
...
                 break;
             case "Night":
-                layer = ArcGISRuntimeEnvironment.createObject("ArcGISVectorTiledLayer", {url:"http://www.arcgis.com/home/item.html?id=bf79e422e9454565ae0cbe9553cf6471"});
+                layer = ArcGISRuntimeEnvironment.createObject("ArcGISVectorTiledLayer", {url:"http://www.arcgis.com/home/item.html?id=92c551c9f07b4147846aae273e822714"});
                 break;
             case "Dark Gray":
-                layer = ArcGISRuntimeEnvironment.createObject("ArcGISVectorTiledLayer", {url:"http://www.arcgis.com/home/item.html?id=850db44b9eb845d3bd42b19e8aa7a024"});
+                layer = ArcGISRuntimeEnvironment.createObject("ArcGISVectorTiledLayer", {url:"http://www.arcgis.com/home/item.html?id=5ad3948260a147a993ef4865e3fad476"});
                 break;
             }
```
lsmallwood: same comment as in C++ - is it a problem if layer is not initialised?

PR #502, opened 2017-04-25
Nand6028/use layer view state status flags

nandinirao:
Assigned to @ldanzinger 

Updating the sample to use statusFlags instead of the obsolete status property.


ldanzinger:
Thank you, looks good

PR #511, opened 2017-05-04
Ldanzinger/raster function fix

ldanzinger:
adding additional null checks, adding the offline table to the doc, and renaming duplicate variable name in QML.

@nandinirao  can you please review/merge

cc @bsnider

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Layers/RasterFunctionFile/RasterFunctionFile.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Layers/RasterFunctionFile/RasterFunctionFile.cpp
@@ -24,6 +24,7 @@
 #include "RasterFunction.h"
 #include "Envelope.h"
 #include <QQmlProperty>
+#include <QFileInfo>
 
 using namespace Esri::ArcGISRuntime;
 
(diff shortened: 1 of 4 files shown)


ArcGISRuntimeSDKQt_CppSamples/Layers/RasterFunctionFile/RasterFunctionFile.cpp
```diff
@@ -89,14 +93,15 @@ RasterFunction* RasterFunctionFile::createRasterFunction()
   // create a RasterFunction
   RasterFunction* rasterFunction = new RasterFunction(m_dataPath + "/color.json");
 
+  // check for valid raster function
+  if (!rasterFunction)
```
nandinirao: Will it ever fail this check? If the path is incorrect, the new will still pass. Ok to keep it like this but its not going to satisfy this condition if the file doesn't exist.

ArcGISRuntimeSDKQt_QMLSamples/Layers/RasterFunctionFile/RasterFunctionFile.qml
```diff
@@ -109,28 +109,35 @@ Rectangle {
...
-        mapView.map.operationalLayers.append(rasterLayer);
+        mapView.map.operationalLayers.append(newRasterLayer);
     }
 
     function createRasterFunction() {
+        // create the raster function
         var rasterFunction = ArcGISRuntimeEnvironment.createObject("RasterFunction", {path: dataPath + "/color.json"});
 
+        // check for valid raster function
+        if (!rasterFunction)
```
nandinirao: Same comment here. Unless we added a API specific check for the path validity.

ldanzinger:
I've added some checks to make sure the file exists @nandinirao

PR #515, opened 2017-05-06
raster function service samples

khajra:
Added the raster function samples for an image service service raster.


khajra:
@ldanzinger Please review.

PR #518, opened 2017-05-08
samples for raster function using a service.

khajra:
Samples for imageServiceRaster as an input for RasterFunction.


khajra:
@ldanzinger Please review.

PR #520, opened 2017-05-10
Khajra/isr2

khajra:
Addition of rendering rule samples for Qt and Qml.


khajra:
@ldanzinger Can you please review?

PR #528, opened 2017-05-18
Lsmallwood/animate map

lsmallwood:
@tdunn - please review the update to enable animation of the map in the sample


tdunn:
@lsmallwood - Nice work! Thanks.

PR #530, opened 2017-05-18
fix sample ui on mobile devices

jtrieu:
@khajra please review and merge. this should fix the issue on mobile devices


khajra:
@jtrieu Looks good!

PR #535, opened 2017-05-25
snippet tags added.

khajra:
Added snippet tags for the samples.


khajra:
@tdunn Can you please review?  Thank you!

PR #537, opened 2017-06-05
Ldanzinger/symbolf

ldanzinger:
@lsmallwood please review and merge. I did a search locally to find all that had `style: "`. Fixing enums and using default properties for some symbols and renderers


lsmallwood:
LGTM

PR #547, opened 2017-06-21
Small fixes.

JamesMBallard:
Make sure we don't call last() on an empty list model.
Make sure the builder is available after calling reset.

@khajra please review. This fixes a crash we're seeing in the test.


tdunn:
@JamesMBallard - PR looks good. Thanks for doing this.

PR #592, opened 2017-10-25
Ldanzinger/viewshed

ldanzinger:


ArcGISRuntimeSDKQt_CppSamples/Analysis/ViewshedLocation/ViewshedLocation.cpp
```diff
@@ -0,0 +1,414 @@
...
+{
+  // Set a viewpoint
+  double x = 6.86088;
+  double y = 45.3604;
+  double z = 3582.55;
+  Point point(x, y, z, SpatialReference(4326));
+  double heading = 345;
+  double pitch = 70;
+  double roll = 0;
+  Camera camera(point, heading, pitch, roll);
```
michael-tims: Are these values stored in local variables for easier readability?

ArcGISRuntimeSDKQt_CppSamples/Analysis/ViewshedLocation/ViewshedLocation.cpp
```diff
@@ -0,0 +1,414 @@
...
+    }
+  });
+
+  connect(m_sceneView, &SceneQuickView::mouseReleased, this, [this](QMouseEvent)
+  {
+    m_calculating = false;
+  });
+}
+
+void ViewshedLocation::createViewshedTask(double x, double y)
```
michael-tims: It's not really a "task", so probably just call the method `createViewshed`

ArcGISRuntimeSDKQt_CppSamples/Analysis/ViewshedLocation/ViewshedLocation.cpp
```diff
@@ -0,0 +1,414 @@
...
+  }
+  else
+  {
+    if (m_obstructedColor == obstructedColor)
+      return;
+
+    m_obstructedColor = obstructedColor;
+  }
+
+  emit obstructedColorChanged();
```
michael-tims: The color methods should use the methods like they are static:
`Viewshed::visibleColor()`, `Viewshed::obstructedColor()`, etc...

ArcGISRuntimeSDKQt_CppSamples/Analysis/ViewshedLocation/ViewshedLocation.qml
```diff
@@ -0,0 +1,304 @@
...
+                        onParameterValueChanged: viewshedSample.minDistance = parameterValue;
+                    }
+
+                    ViewshedSlider {
+                        titleText: qsTr("Max Distance (m)")
+                        parameterValue: viewshedSample.maxDistance
+                        minValue: 1
+                        maxValue: 2000
+                        onParameterValueChanged: viewshedSample.maxDistance = parameterValue;
+                    }
```
michael-tims: Optional... Min and Max distance would be nice as a RangeSlider.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ViewshedLocation/ViewshedLocation.qml
```diff
@@ -0,0 +1,304 @@
...
+                        maxValue: 179
+                        onParameterValueChanged: viewshedSample.horizontalAngle = parameterValue;
+                    }
+
+                    ViewshedSlider {
+                        titleText: qsTr("Heading")
+                        parameterValue: viewshedSample.heading
+                        minValue: 1
+                        maxValue: 359
+                        onParameterValueChanged: viewshedSample.heading = parameterValue;
```
michael-tims: Heading would be nice as a 360 dial, but not sure if one exists in Qt Quick Controls

ArcGISRuntimeSDKQt_CppSamples/Analysis/ViewshedLocation/ViewshedLocation.cpp
ldanzinger: Yes, we are trying to be more explicit in our samples. If you pass things into a function or constructor, and they take the same type for several arguments, it is ambiguous to a reader as to what the value is for, so they then would need to go to the doc to understand the code

michael-tims: Make sense 👍

PR #615, opened 2017-11-17
zoom to extent of the overlay

lsmallwood:
@ldanzinger please review the updates to the sample


ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/GODictionaryRenderer/GODictionaryRenderer.cpp
```diff
@@ -67,12 +71,16 @@ void GODictionaryRenderer::componentComplete()
...
-  m_mapView->graphicsOverlays()->append(m_graphicsOverlay);
 
   parseXmlFile();
-  emit graphicsLoaded();
-  zoomToGraphics();
+  m_mapView->graphicsOverlays()->append(m_graphicsOverlay);
+
+  m_mapView->setMap(m_map);
+
+  connect(m_map, &Map::doneLoading, this, [this]()
```
ldanzinger: we might want to connect before setting the map on the mapview, which will trigger an implicit load. otherwise it is possible that it loads before the connection is made

PR #659, opened 2018-02-06
Updates to ChangeBasemap and VectorTiledLayerUrl samples to point to …

eric906:
Updated samples to point to different vector tile layers


eric906:
@ldanzinger yeah, thanks for reminding me. I updated QML and C++ shots for VectorTiledLayerURL.

PR #669, opened 2018-03-07
Use new AsInt properties

JamesMBallard:
Assign to @lsmallwood 

Updating some samples to use the new "AsInt" properties for assignments. I tested and found no issues. This is the correct thing to do since `typeof identifyLayerResult.geoElements[0].attributes.attributeValue("objectid")` returns "number" and not "string". 

Merging to v.next.


JamesMBallard:
@lsmallwood I don't think so. I expect that will be an incredibly rare event. In my testing of these the values weren't even close to facing truncation.

PR #742, opened 2018-07-25
removing Camera Controller

ldanzinger:
@michael-tims please review/merge. Removing the camera controller per Paul's advice and replacing with an initial viewpoint

Also updating to use a combo box instead of a toggle button so it is clearer to the user about what they are currently looking at

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Scenes/FeatureLayerExtrusion/FeatureLayerExtrusion.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Scenes/FeatureLayerExtrusion/FeatureLayerExtrusion.cpp
@@ -21,7 +21,6 @@
 #include "ArcGISTiledElevationSource.h"
 #include "Scene.h"
 #include "SceneQuickView.h"
-#include "OrbitLocationCameraController.h"
 #include "Point.h"
 #include "SimpleLineSymbol.h"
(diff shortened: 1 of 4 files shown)


ArcGISRuntimeSDKQt_CppSamples/Scenes/FeatureLayerExtrusion/FeatureLayerExtrusion.cpp
```diff
@@ -79,32 +78,35 @@ void FeatureLayerExtrusion::componentComplete()
...
   scene->setBaseSurface(surface);
   scene->operationalLayers()->append(m_featureLayer);
-  m_sceneView->setArcGISScene(scene);
 
-  const Point m_lookAtPoint = Point(-10974490, 4814376, 0, SpatialReference::webMercator());
-  const double distance = 20000000;
-  OrbitLocationCameraController* m_orbitController = new OrbitLocationCameraController(m_lookAtPoint, distance, this);
+  // set initial viewpoint
+  const double distance = 12940924;
+  const Point lookAtPoint = Point(-99.659448, 20.513652, distance, SpatialReference::wgs84());
```
michael-tims: Change to `const Point lookAtPoint(-99.659448, 20.513652, distance, SpatialReference::wgs84());`

ldanzinger:
@michael-tims updated

PR #765, opened 2018-09-11
Fixed ListRelatedFeatures crash.

anmacdonald:
A more correct version of the previous commit, based on feedback from lsmallwood.

Conceptual relationship model had issues.
Duplicate connections were made to the
same table for each layer. One connection
would clean up the contents,
and the remaining connections would attempt to
act upon the corrputed memory.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Features/ListRelatedFeatures/ListRelatedFeatures.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Features/ListRelatedFeatures/ListRelatedFeatures.cpp
@@ -36,6 +36,17 @@
 
 using namespace Esri::ArcGISRuntime;
 
+namespace
+{
+  // Convenience RAII struct that deletes all pointers in given container.
+  struct FeatureQueryListResultLock
+  {
+    FeatureQueryListResultLock(const QList<RelatedFeatureQueryResult*>& list) : results(list) { }
+    ~FeatureQueryListResultLock() { qDeleteAll(results); }
+    const QList<RelatedFeatureQueryResult*>& results;
+  };
+}
+
 ListRelatedFeatures::ListRelatedFeatures(QQuickItem* parent /* = nullptr */):
   QQuickItem(parent)
 {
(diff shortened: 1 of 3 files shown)


ArcGISRuntimeSDKQt_CppSamples/Features/ListRelatedFeatures/ListRelatedFeatures.cpp
```diff
@@ -80,58 +91,56 @@ void ListRelatedFeatures::connectSignals()
...
       {
         m_alaskaNationalParks = static_cast<FeatureLayer*>(m_map->operationalLayers()->at(i));
+        m_alaskaFeatureTable = static_cast<ArcGISFeatureTable*>(m_alaskaNationalParks->featureTable());
+
+        // connect to queryRelatedFeaturesCompleted signal
+        connect(m_alaskaFeatureTable, &ArcGISFeatureTable::queryRelatedFeaturesCompleted,
+                this, [this](QUuid, QList<RelatedFeatureQueryResult*> rawRelatedResults)
+                {
+                  FeatureQueryListResultLock lock(rawRelatedResults);
+                  for (const RelatedFeatureQueryResult* relatedResult : lock.results)
```
lsmallwood: I guess we don't need `qAsConst` here because the member is by const ref?

ArcGISRuntimeSDKQt_CppSamples/Features/ListRelatedFeatures/ListRelatedFeatures.cpp
```diff
@@ -80,58 +91,56 @@ void ListRelatedFeatures::connectSignals()
   {
     if (!loadError.isEmpty())
       return;
-    for (int i = 0; i < m_map->operationalLayers()->size(); i++)
+
+    for (int i = 0; i < m_map->operationalLayers()->size(); ++i)
```
lsmallwood: it looks like this for loop should `break` once it has found the alaska layer?

ArcGISRuntimeSDKQt_CppSamples/Features/ListRelatedFeatures/ListRelatedFeatures.cpp
```diff
@@ -80,58 +91,56 @@ void ListRelatedFeatures::connectSignals()
...
+                  emit showAttributeTable();
+                });
 
         // connect to selectFeaturesCompleted signal
         connect(m_alaskaNationalParks, &FeatureLayer::selectFeaturesCompleted, this, [this](QUuid, FeatureQueryResult* rawResult)
         {
-          QPointer<FeatureQueryResult> result(rawResult);
+          QScopedPointer<FeatureQueryResult> result(rawResult);
           // iterate over features returned
           while (result->iterator().hasNext())
```
lsmallwood: It's kind of weird this is a while loop. I presume there is usually only 1... but the way its written we would query for related features and then zoom to each feature in turn.

ArcGISRuntimeSDKQt_CppSamples/Features/ListRelatedFeatures/ListRelatedFeatures.cpp
anmacdonald: Agreed

PR #796, opened 2018-11-09
Fixed binding loop model crash.

anmacdonald:
``subLayerVisibility`` and checked were feeding off each other in a binding loop that crashed the QML app. Seems to have only manifested with the introduction of QuickControls 2.


lsmallwood:
@JamesMBallard - see https://esri-runtime.slack.com/archives/C0MJH5P5Y/p1541766329001000 for a description of this problem. We can fix this specific crash by changing the likelihood of getting the same object in qml while the unique ptr exists - but it's' not fixing the underlying issue.

PR #805, opened 2018-11-16
EditFeatureAttachments now working

anmacdonald:
EditFeatureAttachments was throwing a syntax error due to missing ArcGISExtras. Removed FileInfo and created an alternative method for parsing filename from URL. Tested on forward-slash and back-slash based systems. Works on both.


lsmallwood:
@anmacdonald - i think we should work out what the problem with `ArcGISExtras` is rather than working around it

PR #808, opened 2018-11-20
FormatCoordinates no longer inherits from a QtQuickItem.

anmacdonald:
Demonstration of how the new model based mapView access will work.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Geometry/FormatCoordinates/FormatCoordinates.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Geometry/FormatCoordinates/FormatCoordinates.cpp
@@ -26,8 +26,8 @@
 
 using namespace Esri::ArcGISRuntime;
 
-FormatCoordinates::FormatCoordinates(QQuickItem* parent) :
-  QQuickItem(parent)
+FormatCoordinates::FormatCoordinates(QObject* parent) :
+  QObject(parent)
 {
 }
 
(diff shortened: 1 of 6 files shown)


ArcGISRuntimeSDKQt_CppSamples/Geometry/FormatCoordinates/FormatCoordinates.cpp
```diff
@@ -177,3 +144,34 @@ QString FormatCoordinates::strUtm() const
 {
   return tr("Utm");
 }
+
+void FormatCoordinates::setMapView(MapQuickView* mapView)
+{
+  // find QML MapView component
+  m_mapView = mapView;
+
+  // create a new basemap instance
```
ldanzinger: - Would it be cleaner to still do much of this work in the constructor? And keep this setter specific to setting the MapView and possibly setting the map on the mapview, graphics overlay on the mapview, etc? 
- Could connectSignals be called from the main constructor instead of here?

ArcGISRuntimeSDKQt_CppSamples/Geometry/FormatCoordinates/main.qml
```diff
@@ -0,0 +1,11 @@
+import QtQuick.Controls 2.2
```
ldanzinger: copyright missing

ArcGISRuntimeSDKQt_CppSamples/Geometry/FormatCoordinates/main.cpp
```diff
@@ -54,17 +53,15 @@ int main(int argc, char *argv[])
...
   // Add the import Path
-  view.engine()->addImportPath(QDir(QCoreApplication::applicationDirPath()).filePath("qml"));
+  engine.addImportPath(QDir(QCoreApplication::applicationDirPath()).filePath("qml"));
   // Add the Runtime and Extras path
   // Add the Runtime and Extras path
-  view.engine()->addImportPath(arcGISRuntimeImportPath);
+  engine.addImportPath(arcGISRuntimeImportPath);
   // Add the Toolkit path
-  view.engine()->addImportPath(arcGISToolkitImportPath);
+  engine.addImportPath(arcGISToolkitImportPath);
```
lsmallwood: I don't this sample actually needs to import the toolkit?

PR #820, opened 2019-01-28
updating anchoring so the delegate content displays in the vertical c…

ldanzinger:
updating anchoring so the delegate content displays in the vertical center and doesn't get cut off

@anmacdonald please review/merge


ldanzinger:
@anmacdonald in this case, the Row is part of the delegate in the ListView. The Column is just a Text with the ListView below it. Not sure it is worth it to use GridLayout in this case since they aren't intermingled really.

PR #824, opened 2019-02-04
Ldanzinger/mspk

ldanzinger:
@lsmallwood please review the code. Not quite ready to merge yet, as I need to update the item id and online download dependencies.


ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+OpenMobileScenePackage::~OpenMobileScenePackage() = default;
+
+// helper method to get cross platform data path
+namespace {
+  QString defaultHomePath()
+  {
+    QString homePath;
+
+  #ifdef Q_OS_ANDROID
```
anmacdonald: This counts for this and the point cloud sample - but you can break down all these methods for a writable location into two (provided without testing):

```
QString defaultHomePath()
{
#if defined(Q_OS_ANDROID) || defined(Q_OS_IOS)
  return QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation);
#else 
  return QStandardPaths::writableLocation(QStandardPaths::HomeLocation);
#endif
}
```

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,150 @@
...
+void OpenMobileScenePackage::packageLoaded(Error e)
+{
+  if (!e.isEmpty())
+  {
+    qDebug() << QString("Package load error: %1 %2").arg(e.message(), e.additionalMessage());
+    return;
+  }
+
+  // obtain the first scene in the list of scenes
+  Scene* scene = m_scenePackage->scenes().at(0);
```
anmacdonald: May crash ungracefully without a check of the size of the scenes struct.

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
anmacdonald: I would also suggest not calling `defaultHomePath` as this is not a home path on devices.

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,150 @@
...
+  {
+    // if direct read is supported, load the package
+    if (supported)
+    {
+      createScenePackage(dataPath);
+    }
+    // otherwise, the package needs to be unpacked
+    else
+    {
+      m_unpackTempPath = QDir::tempPath() + QString("/MspkCpp_%1.mspk").arg(QDateTime().currentMSecsSinceEpoch());
```
anmacdonald: `QStandardLocations::TempLocation` or `QTemporaryDir` would be preferable here.

anmacdonald: If the intention is the unpacked data only exists for the duration of the sample lifetime, I would encourage `QTemporaryDIr`, it would also do away with the need for making the sample name unique with epoch time!

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
ldanzinger: I'm up for the rename, but I think the path refactoring might be a bigger change than I want to do right now. We have a consistent story about where data gets downloaded, and that is baked into the json and is used by the viewer. This seems like it would minimally change the story for Android. This new approach at least improves on what we did before by excluding the Extras dependency for C++. Seems like there is further room for refinement

anmacdonald: I think these are the same directory on Android (`/sdcard` and the user directory `/storage/emulated/0/`).

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -100,7 +100,7 @@ void OpenMobileScenePackage::openPackage()
     // otherwise, the package needs to be unpacked
     else
     {
-      m_unpackTempPath = QDir::tempPath() + QString("/MspkCpp_%1.mspk").arg(QDateTime().currentMSecsSinceEpoch());
+      m_unpackTempPath = QTemporaryDir().path();
```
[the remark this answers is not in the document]
ldanzinger: Seems to work though, as I save it to a member and use that later. is there a better way to approach you'd recommend?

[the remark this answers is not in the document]
anmacdonald: If the `QTemporaryDir` is a member variable, then when it deleted the folder it generates and its contents is gone too. I would recommend doing that instead!

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,153 @@
+// [WriteFile Name=OpenMobileScenePackage, Category=Scenes]
+// [Legal]
+// Copyright 2018 Esri.
```
lsmallwood: 2019

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,153 @@
...
+    dataPath = "/sdcard";
+  #elif defined Q_OS_IOS
+    dataPath = QStandardPaths::writableLocation(QStandardPaths::DocumentsLocation);
+  #else
+    dataPath = QDir::homePath();
+  #endif
+
+    return dataPath;
+  }
+}
```
lsmallwood: add a comment showing this is the end of the anonymous namespace:

`} // namespace`

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+OpenMobileScenePackage::OpenMobileScenePackage(QObject* parent /* = nullptr */):
+  QObject(parent)
+{  
+}
+
+OpenMobileScenePackage::~OpenMobileScenePackage() = default;
+
+// helper method to get cross platform data path
+namespace {
```
lsmallwood: this block (everything in the anonymous namespace) should move up above the c'tor of the class

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,153 @@
...
+  m_sceneView = sceneView;
+
+  openPackage();
+
+  emit sceneViewChanged();
+}
+
+// Open a mobile scene package
+void OpenMobileScenePackage::openPackage()
+{
```
lsmallwood: optional, but all of this logic can be kicked off in the c'tor of the class rather than waiting for the sceneView. Then whenever each of those code paths complete (e.g. when the scene loads or the scene view is set) you wold do:

```
if (m_sceneView && m_scene)
{
   m_sceneView->setArcGISScene(m_scene);
}
```

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,153 @@
...
+  // create the MSPK data path
+  const QString dataPath = defaultDataPath() + "/ArcGIS/Runtime/Data/mspk/philadelphia.mspk";
```
lsmallwood: I think in samples like this one where the data must be downloaded, it would be nice to add a comment to that effect for users who are running these stand-alone

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  // connect to the Mobile Scene Package to determine if direct read is supported
+  connect(MobileScenePackage::instance(), &MobileScenePackage::isDirectReadSupportedCompleted,
```
lsmallwood: maybe add a comment to say something like:

"packages which contain raster data cannot be read directly and must be unpacked first."

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,153 @@
...
+    }
+    // otherwise, the package needs to be unpacked
+    else
+    {
+      m_unpackTempPath = QTemporaryDir().path();
+      MobileScenePackage::unpack(dataPath, m_unpackTempPath);
+    }
+  });
+
+  // connect to the Mobile Scene Package to know when the data is unpacked
```
lsmallwood: "connect to the Mobile Scene Package instance..."

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,153 @@
...
+  emit sceneViewChanged();
+}
+
+// Open a mobile scene package
+void OpenMobileScenePackage::openPackage()
+{
+  // create the MSPK data path
+  const QString dataPath = defaultDataPath() + "/ArcGIS/Runtime/Data/mspk/philadelphia.mspk";
+
+  // connect to the Mobile Scene Package to determine if direct read is supported
```
lsmallwood: "connect to the Mobile Scene Package instance..."

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,153 @@
...
+    }
+    // log that the upack failed
+    else
+    {
+      qDebug() << "failed to unpack";
+    }
+  });
+
+  // Check if the MSPK can be read directly
+  MobileScenePackage::isDirectReadSupported(dataPath);
```
lsmallwood: If we are going to promote the use of the static instance methods, then we also need to connect to the error signal on the instance. I'm going to draft up the code to use the "load and react to the error approach" which I think will be a lot cleaner

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,153 @@
...
+// Slot for handling when the package loads
+void OpenMobileScenePackage::packageLoaded(Error e)
+{
+  if (!e.isEmpty())
+  {
+    qDebug() << QString("Package load error: %1 %2").arg(e.message(), e.additionalMessage());
+    return;
+  }
+
+  if (m_scenePackage->scenes().length() < 1)
```
lsmallwood: I think you can check `if (m_scenePackage->scenes().isEmpty())` instead

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp
```diff
@@ -0,0 +1,153 @@
...
+    return;
+
+  // obtain the first scene in the list of scenes
```
lsmallwood: add a comment to explain that the package could have multiple scenes to choose from - and that the user could be prompted to do that in UI?

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.pro
```diff
@@ -0,0 +1,63 @@
+#-------------------------------------------------
+# Copyright 2018 Esri.
```
lsmallwood: 2019 (throughout)

ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/README.md
```diff
@@ -0,0 +1,35 @@
...
+Opens and displays a scene from a Mobile Scene Package (.mspk).
+
+![](screenshot.png)
+
+## Use Case
+An .mspk file is an archive containing the data (specifically, basemaps and features), used to display an offline 3D scene.
+
+## How it works
+1. Create a `MobileScenePackage` using the path to the local `.mspk` file.
+2. Use `MobileScenePackage::isDirectReadSupported` to check whether the package can be read in the archived form (.mspk) or whether it needs to be unpacked.
```
lsmallwood: these steps are not correct: currently you don't create the package until you have done the isDirectReadSupported check - but as discussed, maybe we can do load and handle errors instead

ldanzinger: ok. this is from the common design, so we probably need to get that updated in that case.

PR #825, opened 2019-02-05
Handle error code approach

lsmallwood:
@ldanzinger, @anmacdonald please review the alternative approach for handling mspk unpack scenario.

I prefer this approach because I think the code is simpler and you don't _always_ incur the cost of the static unpack - which in theory you won't need in the majority of cases.


ldanzinger:
@lsmallwood we discussed this, and I think we all agreed that we can close this for now. Ultimately once we have a more established pattern across the Runtime APIs for something like this, we can promote it, but for now we will use the static async functions to check if valid.

lsmallwood:
yep - let's close for now

PR #828, opened 2019-02-15
Jcaccamo/ranged based for loops

DonMoJa:
This PR updates our samples from the long handed c-style for loops to ranged based for loops where applicable.  

@anmacdonald @ldanzinger - Could y'all (trying to keep fresh with my recently acquired southern drawl) review when you have some time? Thank you.


ArcGISRuntimeSDKQt_CppSamples/Analysis/StatisticalQuery/StatisticalQuery.cpp
```diff
@@ -89,9 +89,9 @@ void StatisticalQuery::componentComplete()
     {
       StatisticRecord* record = iter.next(&parent);
       const QVariantMap& statsMap = record->statistics();
-      for (auto it = statsMap.cbegin(); it != statsMap.cend(); ++it)
+      for (const auto& it : statsMap.toStdMap())
```
anmacdonald: As far as I am aware `QVariantMap` has an iterator type. Why are we converting in this instance? This goes foe the other instances too.

ArcGISRuntimeSDKQt_CppSamples/Analysis/StatisticalQueryGroupSort/StatisticalQueryGroupSort.cpp
```diff
@@ -106,17 +106,17 @@ void StatisticalQueryGroupSort::connectSignals()
...
       }
       const QString sectionString = sectionStrings.join(',');
 
       // obtain the statistics
       const QVariantMap& statsMap = record->statistics();
-      for (auto it = statsMap.cbegin(); it != statsMap.cend(); ++it)
+      for (const auto& it : statsMap.toStdMap())
       {
-        QString statString = QString("%1: %2").arg(it.key(), it.value().toString());
+        QString statString = QString("%1: %2").arg(it.first, it.second.toString());
```
anmacdonald: Add const here please!

ArcGISRuntimeSDKQt_CppSamples/Analysis/StatisticalQuery/StatisticalQuery.cpp
anmacdonald: Wait I see - the iterator dereference is the value type only, and you need key/value pairs. I wonder if this change is appropriate in this case.

DonMoJa: Oh I think you are right and that this was only supposed to apply to list model types and I just got carried away.

ArcGISRuntimeSDKQt_CppSamples/Analysis/StatisticalQueryGroupSort/StatisticalQueryGroupSort.cpp
DonMoJa: Good catch! Thanks for pointing that out.

ArcGISRuntimeSDKQt_CppSamples/Analysis/StatisticalQuery/StatisticalQuery.cpp
ldanzinger: the main purpose of the issue was to scan the repo to replace any c style for loops with range based for loops for list model types, since that is the new support James recently added. Looks like maybe we didn't have any instances of this in the samples? If there are other improvements to make, I'm ok with having those go in. I'll defer to @anmacdonald on the review comments

JamesMBallard: I would back out this change. Converting this to a std::map should not be needed.

DonMoJa: Thanks for the input! I will back this up to what was there prior.

ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap/GenerateOfflineMap.cpp
```diff
@@ -126,9 +126,9 @@ void GenerateOfflineMap::generateMapByExtent(double xCorner1, double yCorner1, d
           {
             QString layerErrors = "";
             const QMap<Layer*, Error>& layerErrorsMap = generateJob->result()->layerErrors();
-            for (auto it = layerErrorsMap.cbegin(); it != layerErrorsMap.cend(); ++it)
+            for (const auto& it : layerErrorsMap.toStdMap())
```
anmacdonald: Additional `toStdMap`

ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap_Overrides/GenerateOfflineMap_Overrides.cpp
```diff
@@ -340,9 +340,9 @@ void GenerateOfflineMap_Overrides::takeMapOffline(const QString& dataPath)
       {
         QString layerErrors = "";
         const QMap<Layer*, Error>& layerErrorsMap = generateJob->result()->layerErrors();
-        for (auto it = layerErrorsMap.cbegin(); it != layerErrorsMap.cend(); ++it)
+        for (const auto& it : layerErrorsMap.toStdMap())
```
anmacdonald: Additional `stdMap`

DonMoJa: Thank you for pointing these two out!

PR #857, opened 2019-04-11
Ldanzinger/grouplayers

ldanzinger:
@JamesMBallard please review/merge


ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/GroupLayers.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+  // create initial viewpoint
+  Geometry env = Geometry::fromJson(R"({"spatialReference":{
+                                    "wkid":4326
+                                    },
+                                    "xmax":-122.67960721754773,
+                                    "xmin":-122.68647066116789,
+                                    "ymax":45.53584958588318,
+                                    "ymin":45.531539857343745
+                                    })");
```
JamesMBallard: Any reason this is needed over just constructing an Envelope here?

ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/GroupLayers.cpp
```diff
@@ -0,0 +1,121 @@
...
+  // zoom to viewpoint
+  m_sceneView->setViewpoint(Viewpoint(env.extent()));
```
JamesMBallard: Also, if you have an Envelope to begin with then `env.extent()` is redundant.

ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/GroupLayers.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+  emit sceneViewChanged();
+}
+
+LayerListModel* GroupLayers::getGroupLayerListModel(int layerId)
+{
+  if (!m_scene)
+    return nullptr;
+
+  auto groupLayer = static_cast<GroupLayer*>(m_scene->operationalLayers()->at(layerId));
```
JamesMBallard: Please use the actual type name for samples.

`GroupLayer* groupLayer = ...`

ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/GroupLayers.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+  if (groupLayer)
+    return groupLayer->layers();
+  else
+    return nullptr;
```
JamesMBallard: This block can be reformatted a bit for readability. IMO you don't need the `else` case.

```c++
  if (groupLayer)
    return groupLayer->layers();

  return nullptr;
```

ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/GroupLayers.h
```diff
@@ -0,0 +1,64 @@
...
+class QAbstractListModel;
+
+#include <QObject>
+
+class GroupLayers : public QObject
+{
+  Q_OBJECT
+
+  Q_PROPERTY(Esri::ArcGISRuntime::SceneQuickView* sceneView READ sceneView WRITE setSceneView NOTIFY sceneViewChanged)
+  Q_PROPERTY(Esri::ArcGISRuntime::LayerListModel* layerListModel READ layerListModel NOTIFY layerListModelChanged)
```
JamesMBallard: Possible cleanup: use `MEMBER` syntax here and you can get rid of the getter below entirely. I've found that works quite well for read-only properties with trivial getters.

```C++
Q_PROPERTY(Esri::ArcGISRuntime::LayerListModel* layerListModel MEMBER m_layerListModel NOTIFY layerListModelChanged)
```

ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/GroupLayers.h
```diff
@@ -0,0 +1,64 @@
...
+  void sceneViewChanged();
+  void layerListModelChanged();
+
+public:
+  Q_INVOKABLE Esri::ArcGISRuntime::LayerListModel* getGroupLayerListModel(int layerId);
+
+private:
+  Esri::ArcGISRuntime::SceneQuickView* sceneView() const;
+  void setSceneView(Esri::ArcGISRuntime::SceneQuickView* sceneView);
+  Esri::ArcGISRuntime::LayerListModel* layerListModel() const { return m_layerListModel; }
```
JamesMBallard: If you use `MEMBER` you don't need this anymore. If you want to keep this, move the implementation to the .cpp file.

ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/GroupLayers.qml
```diff
@@ -0,0 +1,132 @@
...
+                model: sampleModel.layerListModel
+
+                // Assign the delegate to the delegate created above
+                delegate: Item {
+                    height: childrenRect.height
+
+                    // select the component based on the layer type
+                    Loader {
+                        sourceComponent: layerType === 22 ?
+                                             groupLayerDelegate : layerDelegate
```
JamesMBallard: Is there any simple way to avoid exposing this as an int? `layerType === 22` requires a bit of explanation. Maybe expose a single-value enum for this sample that only contains the GroupLayer layer type value? If that's overkill or more confusing then this is fine.

ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/README.md
```diff
@@ -0,0 +1,38 @@
...
+In a land development project, you might group layers according to the phase of development.
+
+## How to use the sample
+
+The layers in the map will be displayed in a table of contents. Toggle the checkbox next to a layer's name to change its visibility. Turning a group layer's visibility off will override the visibility of its child layers.
+
+## How it works
+
+1. Create an empty `GroupLayer`.
+2. Add a child layer to the group layer's layers collection.
```
JamesMBallard: Should this explanation go the other way around? It's not a child layer to begin with since you're building up the group layer from scratch.

Maybe: `Add the new layer to the group layer's layers collection so it will be a child layer.`, or something like that?

ArcGISRuntimeSDKQt_QMLSamples/Layers/GroupLayers/GroupLayers.qml
```diff
@@ -0,0 +1,189 @@
...
+                                                                json: {
+                                                                    "spatialReference": {
+                                                                        "wkid":4326
+                                                                    },
+                                                                    "xmax":-122.67960721754773,
+                                                                    "xmin":-122.68647066116789,
+                                                                    "ymax":45.53584958588318,
+                                                                    "ymin":45.531539857343745
+                                                                }
+                                                            });
```
JamesMBallard: Same question from C++. Do you need to use json here specifically? It's more efficient to create the object by assigning the properties directly.

ArcGISRuntimeSDKQt_QMLSamples/Layers/GroupLayers/README.md
```diff
@@ -0,0 +1,38 @@
...
+In a land development project, you might group layers according to the phase of development.
+
+## How to use the sample
+
+The layers in the map will be displayed in a table of contents. Toggle the checkbox next to a layer's name to change its visibility. Turning a group layer's visibility off will override the visibility of its child layers.
+
+## How it works
+
+1. Create an empty `GroupLayer`.
+2. Add a child layer to the group layer's layers collection.
```
JamesMBallard: Same comment as previous description.  This one may make sense as-is since the layers are nested in place within the GroupLayer object for Qml.

ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/GroupLayers.cpp
ldanzinger: Not particularly. Just showcasing another valid way to create geometries that we don't show very often in public code. I think it's ok. Do you prefer it be changed?

ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/GroupLayers.qml
ldanzinger: I was thinking about that. Not sure if exposing a new enum is clearer. Not really a great way to support this with our C++ API. Maybe I'll just add a comment explaining

ldanzinger:
thanks @JamesMBallard. Comments addressed except the Geometry::fromJson and the enum from C++ to QML. I can change the Geometry one if you feel it is better. Not sure if there is really a great solution for the enums

ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/GroupLayers.cpp
JamesMBallard: Sure, this is fine if it's showing an alternate method folks may not know about.

ArcGISRuntimeSDKQt_CppSamples/Layers/GroupLayers/GroupLayers.qml
JamesMBallard: That sounds ok

PR #863, opened 2019-04-16
Fix the iOS build.

JamesMBallard:
Fixing the iOS build. If this sample gets ported to master for current release, make sure this fix gets ported as well.

cc: @ecmorris


JamesMBallard:
Assigned to @ldanzinger. Please review.

PR #868, opened 2019-04-22
Ldanzinger/fixes from verifications

ldanzinger:
@JamesMBallard please review/merge. A couple of issues addressed:

- setting width on text so elide works if text is too long
- Kathleen has reported random crash on iOS when it starts. Appears to most likely be a timing issue of some sort. I could repro one time only, making it hard to pin point. Adding 2 speculative safety measures that could help:
  - setting initial value to symbol size and color in case the block to set these on the symbol executes before the combo box values are initialized
  - checking casted pointer to ensure cast was successful

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ReadSymbolsFromMobileStyle/ReadSymbolsFromMobileStyle.h
+++ b/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ReadSymbolsFromMobileStyle/ReadSymbolsFromMobileStyle.h
@@ -79,8 +79,8 @@ private:
   SymbolImageProvider* m_symbolImageProvider = nullptr;
   QList<Esri::ArcGISRuntime::SymbolStyleSearchResultListModel*> m_models = { nullptr, nullptr, nullptr, nullptr };
   QList<QUuid> m_taskIds;
-  QColor m_currentColor;
-  int m_symbolSize;
+  QColor m_currentColor = QColor("yellow");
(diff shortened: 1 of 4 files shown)


ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ReadSymbolsFromMobileStyle/ReadSymbolsFromMobileStyle.h
```diff
@@ -79,8 +79,8 @@ class ReadSymbolsFromMobileStyle : public QObject
   SymbolImageProvider* m_symbolImageProvider = nullptr;
   QList<Esri::ArcGISRuntime::SymbolStyleSearchResultListModel*> m_models = { nullptr, nullptr, nullptr, nullptr };
   QList<QUuid> m_taskIds;
-  QColor m_currentColor;
-  int m_symbolSize;
+  QColor m_currentColor = QColor("yellow");
```
JamesMBallard: Use the enum color constructor. This is one of the clazy warnings that will come up. 

Be careful, however. Make sure that the "yellow" string corresponds with `Qt::yellow` enum. That is not is obvious is it should be.
`QColor m_currentColor = QColor(Qt::yellow)`

ldanzinger: seems to work well. thanks

PR #869, opened 2019-04-22
initial commit of wfs with xml query sample

ldanzinger:
@anmacdonald please review


ArcGISRuntimeSDKQt_QMLSamples/Layers/LoadWfsXmlQuery/LoadWfsXmlQuery.qml
```diff
@@ -0,0 +1,71 @@
...
+import Esri.ArcGISRuntime 100.6
+
+Rectangle {
+    id: rootRectangle
+    clip: true
+    width: 800
+    height: 600
+
+    readonly property string xmlQueryString :
+     '
```
anmacdonald: Genuinely surprised single ticks work here for multiline strings! Backticks (`) tend to be the accepted standard for multiline strings in Javascript. Not sure if case for QML too?

ArcGISRuntimeSDKQt_CppSamples/Layers/LoadWfsXmlQuery/LoadWfsXmlQuery.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "Map.h"
+#include "MapQuickView.h"
+#include "FeatureLayer.h"
+#include "WfsFeatureTable.h"
+
+using namespace Esri::ArcGISRuntime;
+
+namespace
+{
+  static const QString xmlQueryString = R"(
```
anmacdonald: An unfortunate C++ issue is that `static` has several meanings in different contexts. In this instance `static` and an anonymous namespace are actually the same thing! You should remove the `static` keyword from here.

ArcGISRuntimeSDKQt_QMLSamples/Layers/LoadWfsXmlQuery/LoadWfsXmlQuery.qml
ldanzinger: @anmacdonald yep it works. The JS backtick is new ES6 functionality and the default Qt Creator doesn't yet support it, so I opted to use this instead since it works well. Otherwise it will run but creator will complain about syntax errors OOTB

PR #871, opened 2019-04-23
fixing crash by checking for nullptr

ldanzinger:
@JamesMBallard please review/merge. I was missing one null check which caused the crash. Need to explicitly call update on searchSymbolsCompleted now so that the symbol updates once each component is complete

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ReadSymbolsFromMobileStyle/ReadSymbolsFromMobileStyle.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ReadSymbolsFromMobileStyle/ReadSymbolsFromMobileStyle.cpp
@@ -71,7 +71,7 @@ ReadSymbolsFromMobileStyle::ReadSymbolsFromMobileStyle(QObject* parent /* = null
     m_models[index] = results;
 
     emit symbolResultsChanged();
-    updateSymbol();
+    updateSymbol(0, 0, 0, Qt::yellow, 40);
   });
 
   // Load the style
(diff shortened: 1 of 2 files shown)


ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ReadSymbolsFromMobileStyle/ReadSymbolsFromMobileStyle.h
```diff
@@ -56,7 +56,7 @@ class ReadSymbolsFromMobileStyle : public QObject
   static void init();
 
   Q_INVOKABLE void clearGraphics();
-  Q_INVOKABLE void updateSymbol(int hatIndex, int mouthIndex, int eyeIndex, QColor color, int size);
+  Q_INVOKABLE void updateSymbol(int hatIndex = 0, int mouthIndex = 0, int eyeIndex = 0, QColor color = QColor(Qt::yellow), int size = 40);
```
JamesMBallard: Not generally good practice to have that many default args. Can you just call it with the defaults instead?

ldanzinger: updated

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ReadSymbolsFromMobileStyle/ReadSymbolsFromMobileStyle.cpp
```diff
@@ -71,6 +71,7 @@ ReadSymbolsFromMobileStyle::ReadSymbolsFromMobileStyle(QObject* parent /* = null
     m_models[index] = results;
 
     emit symbolResultsChanged();
+    updateSymbol(0, 0, 0, Qt::yellow, 40);
```
JamesMBallard: This should be `QColor(Qt::yellow)` for readability and to avoid implicit conversion.

ldanzinger: alright, should be updated now. thanks

PR #883, opened 2019-04-29
adding temp path to local server samples

ldanzinger:
@khajra please review/merge. Per Mike B, it is best practice to set temp path for local server, so I am adding this to the local server samples

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/LocalServer/DynamicWorkspaceRaster/DynamicWorkspaceRaster.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/LocalServer/DynamicWorkspaceRaster/DynamicWorkspaceRaster.cpp
@@ -29,6 +29,7 @@
 #include <QDir>
 #include <QFile>
 #include <QFileInfo>
+#include <QTemporaryDir>
 
 using namespace Esri::ArcGISRuntime;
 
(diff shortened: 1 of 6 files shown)


mbranscomb:
@ldanzinger What does `tempDir.path()` resolve to?

mbranscomb:
@ldanzinger Note WPF implemented https://github.com/Esri/arcgis-runtime-samples-dotnet/blob/b9b4ea556ce6130d6d2c783ea613c6ae702c9332/src/WPF/ArcGISRuntime.WPF.Viewer/App.xaml.cs#L24

DonMoJa:
tempDir.path() is resolving to "C:/Users/jare8800/AppData/Local/Temp/1/..." which is created upon setting the LocalServer::setTempDataPath. It seems to be deleted almost immediately after this but I am unsure if this is expected or not.

Also the Qt documentation does state, ["This method may be called when the local server is not started and is not starting"](https://developersdev.arcgis.com/qt/latest/cpp/api-reference/esri-arcgisruntime-localserver.html#setTempDataPath) which in this case in this situation. Although when I tried doing it beforehand it would throw and error consistently stating, "Startup failed: 'dataDirPath' was set to a non-writable path: ...".

Also when looking at the logs of local server it states there was an error with unpacking the package:
image

PR #890, opened 2019-05-15
modifying code so that initial viewpoint is honored

DonMoJa:
Originally created a scene from a portal item but it was not honoring the viewpoint in C++ so made some changes in order for it to display from the viewpoint established by the data.

@JamesMBallard please review when you have a moment. Thank you!

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Scenes/ViewContentBeneathTerrainSurface/ViewContentBeneathTerrainSurface.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Scenes/ViewContentBeneathTerrainSurface/ViewContentBeneathTerrainSurface.cpp
@@ -19,17 +19,15 @@
 #include "ArcGISTiledElevationSource.h"
 #include "Scene.h"
 #include "SceneQuickView.h"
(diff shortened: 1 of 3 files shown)


ArcGISRuntimeSDKQt_CppSamples/Scenes/ViewContentBeneathTerrainSurface/ViewContentBeneathTerrainSurface.cpp
```diff
@@ -19,17 +19,15 @@
 #include "ArcGISTiledElevationSource.h"
 #include "Scene.h"
 #include "SceneQuickView.h"
-#include "Portal.h"
-#include "PortalItem.h"
+#include "SceneView.h"
```
JamesMBallard: Is this needed? `SceneQuickView` inherits `SceneView`.

DonMoJa: Good catch! It definitely is not needed.

PR #892, opened 2019-05-22
setting temp path to explicit location

ldanzinger:
@JamesMBallard please review/merge. Iterating on changes pushed up as part of https://github.com/Esri/arcgis-runtime-samples-qt/pull/883

@mbranscomb has let me know that due to path length issues, we need to explicitly set the temp path for local server samples to something shorter. 

As part of this pr I tried QDir::rootPath() so it was very short, but I think we are likely to hit permission issues especially on Linux. QDir::homePath should more reliably have write permissions I think

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/LocalServer/LocalServerFeatureLayer/LocalServerFeatureLayer.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/LocalServer/LocalServerFeatureLayer/LocalServerFeatureLayer.cpp
@@ -26,7 +26,6 @@
 #include "Viewpoint.h"
 
 #include <QDir>
-#include <QTemporaryDir>
 
 using namespace Esri::ArcGISRuntime;
 
@@ -81,9 +80,15 @@ void LocalServerFeatureLayer::connectSignals()
     {
       qDebug() << "Local server started";
 
-      // set temp path
-      QTemporaryDir tempDir;
-      LocalServer::instance()->setTempDataPath(tempDir.path());
+      // create temp path
+      const QString tempPath = QDir::homePath() + "/EsriQtTemp";
+
+      // create the directory
+      if (!QDir(tempPath).exists())
+        QDir().mkdir(tempPath);
+
+      // set the temp data path for the local server
+      LocalServer::instance()->setTempDataPath(tempPath);
 
       // start the local service
       m_localFeatureService->start();
--- a/ArcGISRuntimeSDKQt_CppSamples/LocalServer/LocalServerFeatureLayer/LocalServerFeatureLayer.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/LocalServer/LocalServerFeatureLayer/LocalServerFeatureLayer.cpp
@@ -79,19 +79,6 @@ void LocalServerFeatureLayer::connectSignals()
     if (LocalServer::status() == LocalServerStatus::Started)
     {
       qDebug() << "Local server started";
-
-      // create temp path
-      const QString tempPath = QDir::homePath() + "/EsriQtTemp";
-
-      // create the directory
-      if (!QDir(tempPath).exists())
-        QDir().mkdir(tempPath);
-
-      // set the temp data path for the local server
-      LocalServer::instance()->setTempDataPath(tempPath);
-
-      // start the local service
-      m_localFeatureService->start();
     }
   });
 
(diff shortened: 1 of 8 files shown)


ArcGISRuntimeSDKQt_CppSamples/LocalServer/LocalServerFeatureLayer/LocalServerFeatureLayer.cpp
```diff
@@ -81,9 +80,15 @@ void LocalServerFeatureLayer::connectSignals()
...
 
-      // set temp path
-      QTemporaryDir tempDir;
-      LocalServer::instance()->setTempDataPath(tempDir.path());
+      // create temp path
+      const QString tempPath = QDir::homePath() + "/EsriQtTemp";
+
+      // create the directory
+      if (!QDir(tempPath).exists())
+        QDir().mkdir(tempPath);
```
JamesMBallard: This applies to all. Rather than creating two QDir temporaries here, would this work?

```c++
QDir dir(tempPath);
if (dir.exists())
  dir.mkdir(tempPath);
```
```

File: ArcGISRuntimeSDKQt_CppSamples/LocalServer/LocalServerMapImageLayer/LocalServerMapImageLayer.cpp [hunk 2]
```diff
@@ -78,9 +77,15 @@ void LocalServerMapImageLayer::connectSignals()
   {
     if (LocalServer::status() == LocalServerStatus::Started)
     {
-      // set temp path
-      QTemporaryDir tempDir;
-      LocalServer::instance()->setTempDataPath(tempDir.path());
+      // create temp path
+      const QString tempPath = QDir::homePath() + "/EsriQtTemp";
```
JamesMBallard: This could be a much longer path now on Linux. `/tmp` is about as short as you can hope for.

ArcGISRuntimeSDKQt_CppSamples/LocalServer/LocalServerServices/LocalServerServices.cpp
```diff
@@ -24,7 +24,6 @@
 
 #include <QDesktopServices>
 #include <QDir>
-#include <QTemporaryDir>
```
JamesMBallard: Not using `QTemporaryDir` now also has the unfortunate side effect of leaving around all these temporary directories. Is there a way to handle that?

mbranscomb:
@ldanzinger Looks like the pattern for Qt is a little different to .NET, where the TempPath property is on the ArcGISRuntimeEnvironment class and must be set prior to ArcGIS Runtime initialization e.g. https://github.com/Esri/arcgis-runtime-samples-dotnet/blob/109375704bac56dd50b91cd4687f39b801a0134d/src/WPF/ArcGISRuntime.WPF.Viewer/App.xaml.cs#L27

...But otherwise your changes look Ok to me.

ArcGISRuntimeSDKQt_CppSamples/LocalServer/LocalServerServices/LocalServerServices.cpp
```diff
@@ -78,8 +78,13 @@ void LocalServerServices::connectSignals()
       }
       case LocalServerStatus::Started:
       {
-        // set temp path
-        QTemporaryDir tempDir;
+        // create temp path
+        const QString tempPath = LocalServerServices::shortestTempPath() + "/EsriQtTemp";
+
+        // create the directory
+        const QTemporaryDir tempDir(tempPath);
```
JamesMBallard: My bad for not noticing this before, but QTemporaryDir  needs to be a class member for this to work (probably a `std::unique_ptr<QTemporaryDir>`). This class deletes the temp dir on destruction, and it will destruct at the end of this block.

ArcGISRuntimeSDKQt_CppSamples/LocalServer/LocalServerGeoprocessing/LocalServerGeoprocessing.cpp
```diff
@@ -100,8 +100,13 @@ void LocalServerGeoprocessing::connectSignals()
   {
     if (LocalServer::status() == LocalServerStatus::Started)
     {
-      // set temp path
-      QTemporaryDir tempDir;
+      // create temp path
+      const QString tempPath = LocalServerGeoprocessing::shortestTempPath() + "/EsriQtTemp";
+
+      // create the directory
+      const QTemporaryDir tempDir(tempPath);
```
JamesMBallard: See other comment about QTemporaryDir.

ldanzinger:
@JamesMBallard should be updated now to set temp and app data paths to a stored temp dir. the temp dir gets cleaned up at the end of the lifecycle of the app

PR #893, opened 2019-05-28
Jcaccamo/browse wfs layers

DonMoJa:
@anmacdonald @mfeigl Please review. Thank you!


ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.cpp
```diff
@@ -0,0 +1,175 @@
...
+void BrowseWfsLayers::createWfsFeatureTable(int index, bool swap)
+{
+  // clear previous layer
+  m_map->operationalLayers()->clear();
+  // set viewpoint to extent of selected layer
+  m_mapView->setViewpointGeometry(m_wfsLayersInfoList[index].extent());
+  // create WFS Feature Table from selected layer
+  m_wfsFeatureTable = new WfsFeatureTable(m_wfsLayersInfoList[index], this);
+
+  // Set feature request mode to manual - only manual is supported at v100.5.
```
mfeigl: Should a comment be left, saying what is supported at v100.6?

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.cpp
```diff
@@ -0,0 +1,175 @@
...
+      for(auto i : m_wfsLayersInfoList)
+        m_layerInfoTitleList.append(i.title());
+
+      emit layerInfoTitleListChanged();
+  });
+
+  m_wfsService->load();
+
+  // set initial viewpoint
+//  Envelope env(-122.341581, 47.613758, -122.332662, 47.617207, SpatialReference::wgs84());
```
mfeigl: Why are these lines commented out? Should they be removed?

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.cpp
DonMoJa: Good question. This is a 100.5 sample so that may be a step too far but I won't take the final say on this. @ldanzinger any thoughts on this?

mfeigl: Yes, that's fair enough. 
May I ask, however, about the version number in BrowseWfsLayers.pro which says `ARCGIS_RUNTIME_VERSION = 100.6`: https://github.com/Esri/arcgis-runtime-samples-qt/pull/893/files#diff-03dda30ae078e83b2627972fb6c63cebR22 should it not say `100.5`?

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.cpp
```diff
@@ -0,0 +1,169 @@
...
+  m_wfsService = new WfsService(QUrl("https://dservices2.arcgis.com/ZQgQTuoyBrtmoGdP/arcgis/services/Seattle_Downtown_Features/WFSServer?service=wfs&request=getcapabilities"),this);
+
+  // once WFS Service is loaded, populate QStringList with layer names for ComboBox
+  connect(m_wfsService, &WfsService::doneLoading, this, [this](Error e)
+  {
+    if (!e.isEmpty())
+      return;
+
+      m_wfsLayersInfoList = m_wfsService->serviceInfo().layerInfos();
+      for(auto i : m_wfsLayersInfoList)
```
anmacdonald: So, as good practice for samples, we try and reduce the amount of `autos` as much as possible to make explicit to the user what is going on. `auto` doesn't automatically add const-ref qualifiers either, so this is doing an implicit copy for every access! Try:

`for (const WfsLayerInfo& i : m_wfsLayersInfoList)`

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.cpp
anmacdonald: This will likely be safe enough - I don't think it will be going into master until 100.6.0.

anmacdonald: Is there a ticket prepared for updating this comment when we support automatic?

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.cpp
```diff
@@ -0,0 +1,169 @@
...
+  m_mapView->setViewpointGeometry(m_wfsLayersInfoList[index].extent());
+  // create WFS Feature Table from selected layer
+  m_wfsFeatureTable = new WfsFeatureTable(m_wfsLayersInfoList[index], this);
+
+  // Set feature request mode to manual - only manual is supported at v100.5.
+  // In this mode, you must manually populate the table - panning and zooming
+  // won't request features automatically.
+  m_wfsFeatureTable->setFeatureRequestMode(FeatureRequestMode::ManualCache);
+
+  if(swap == true)
```
anmacdonald: `if (swap)`

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.cpp
```diff
@@ -0,0 +1,169 @@
...
+    m_wfsFeatureTable->setAxisOrder(OgcAxisOrder::Swap);
+  else
+    m_wfsFeatureTable->setAxisOrder(OgcAxisOrder::NoSwap);
+
+  // once WFS Feature Table is loaded, populate the table and add the layer to the map
+  connect(m_wfsFeatureTable, &WfsFeatureTable::doneLoading, this, [this](Error e){
```
anmacdonald: newline for curly bracket. Either needs aligned with the `connect` or with the start of the capture `[this]`.

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.cpp
```diff
@@ -0,0 +1,169 @@
...
+    return;
+
+  // create query parameters
+  QueryParameters params;
+  params.setGeometry(m_mapView->visibleArea().extent());
+  params.setSpatialRelationship(SpatialRelationship::Intersects);
+
+  // query the service
+  constexpr bool clearCache = false;
+  const QStringList outFields = {"*"};
```
anmacdonald: Equals is not required here as `{"*"}` is the constructor! Try: `const QStringList outFields {"*"};`

Also be sure to test on Linux when applying uniformed initialization. GCC 4.8 is always the compiler that trips up with this C++11 features.

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.cpp
```diff
@@ -0,0 +1,169 @@
...
+  m_wfsFeatureTable->populateFromService(params, clearCache, outFields);
+}
+
+void BrowseWfsLayers::addFeatureLayerToMap()
+{
+  // create feature layer from the feature table
+  FeatureLayer* featureLayer = new FeatureLayer(m_wfsFeatureTable, this);
+  SimpleRenderer* sr;
```
anmacdonald: It may seem redundant, but best practice is:

```
SimpleRenderer* sr = nullptr;
SimpleMarkerSymbol* sms = nullptr;
SimpleLineSymbol* sls = nullptr;
```

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.cpp
```diff
@@ -0,0 +1,169 @@
...
+    break;
+  case GeometryType::Polygon:
+    sls = new SimpleLineSymbol(SimpleLineSymbolStyle::Solid, QColor("blue"), 3.0f, this);
+    sr = new SimpleRenderer(sls, this);
+    break;
+  case GeometryType::Polyline:
+    sls = new SimpleLineSymbol(SimpleLineSymbolStyle::Solid, QColor("red"), 3.0f, this);
+    sr = new SimpleRenderer(sls, this);
+    break;
+  case GeometryType::Unknown:
```
anmacdonald: Maybe worth a token error message just in-case?

```
default:
  qDebug() << "Error! Unsupported GeometryType!"; 
  return;
```

or

```
case GeometryType::Unknown:
case GeometryType::Multipoint:
case GeometryType::Envelope:
  qDebug() << "Error! Unsupported GeometryType!";
  return;
```

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.qml
```diff
@@ -0,0 +1,88 @@
...
+                id: layersComboBox
+                model: browseWfsLayersSampleModel.layerInfoTitleListModel
+                Layout.fillWidth: true
+                Layout.margins: 3
+                Layout.alignment: Qt.AlignHCenter
+            }
+
+            Button {
+                id: loadSelectedLayerBtn
+                text: qsTr("Load Selected Layer")
```
anmacdonald: Elegant QML!

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.qml
```diff
@@ -0,0 +1,88 @@
...
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// [Legal]
+
+import QtQuick 2.6
+import QtQuick.Controls 2.2
```
anmacdonald: Don't need both:

```
import QtQuick.Controls 2.2
import QtQuick.Controls 2.12
```

DonMoJa:
@anmacdonald code has been updated and slightly modified due to some feedback from the samples team.

mfeigl:
Looks good to me too.

PR #895, opened 2019-05-31
Ldanzinger/manage layers

ldanzinger:
@JamesMBallard please review/merge. This one had a few challenges. 

First is the reverse draw order where the first layer will display at the top of the list view, whereas most TOCs have the last added layer at the top, as that one will draw on the top. Have worked around this with a QSortFilterProxyModel.

Second challenge is that this sample maintains 2 lists of layers. One is for the layers in the map and the other is for layers removed from the map. We need to show both lists in 2 separate views. This is dead easy with the LayerListModel. But that is not creatable, so I have to come up with another data structure to store the "removed/deleted" layer list. I wanted to just do a QList<Layer*> and then access on the QML side via modelData, but that would require Q_PROPERTYs to be exposed for things like Layer::name, which we don't have with the C++ API. So I end up exposing the names a QStringList and storing the layer in a QList<Layer*>. It works ok but feels like it could be cleaner. Being able to create another LayerListModel would make it simple, but not sure if there are other use cases where this could be used or if it is sort of contrived.


ArcGISRuntimeSDKQt_CppSamples/Layers/ManageOperationalLayers/DrawOrderLayerListModel.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+int DrawOrderLayerListModel::mappedIndex(int index) const
+{
+  const QModelIndex sourceIndex = mapToSource(this->index(index, 0));
+  return sourceIndex.row();
+}
+
+bool DrawOrderLayerListModel::lessThan(const QModelIndex& sourceLeft, const QModelIndex& sourceRight) const
+{
+  return sourceLeft.row() > sourceRight.row();
```
JamesMBallard: To be a true inverse, should this be `>=` ?

ArcGISRuntimeSDKQt_CppSamples/Layers/ManageOperationalLayers/ManageOperationalLayers.cpp
```diff
@@ -0,0 +1,137 @@
...
+  m_mapView = mapView;
+  m_mapView->setMap(m_map);
+
+  emit mapViewChanged();
+}
+
+QAbstractItemModel* ManageOperationalLayers::layerListModel() const
+{
+  if (!m_drawOrderListModel)
+    return nullptr;
```
JamesMBallard: This check is not needed. `return m_drawOrderListModel;` means the same as `return nullptr` if it is nullptr.

ArcGISRuntimeSDKQt_CppSamples/Layers/ManageOperationalLayers/ManageOperationalLayers.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  return m_drawOrderListModel;
+}
+
+void ManageOperationalLayers::moveLayerUp(int index)
+{
+  if (!m_drawOrderListModel || !m_map->operationalLayers())
```
JamesMBallard: This comment applies throughout, but do you need to check if the operational layers are valid, or just the map? if you have a map, the operational layers list model is always valid.

ArcGISRuntimeSDKQt_CppSamples/Layers/ManageOperationalLayers/ManageOperationalLayers.cpp
```diff
@@ -0,0 +1,137 @@
...
+    return;
+
+  // get the index
+  int modelIndex = m_drawOrderListModel->mappedIndex(index);
+  Layer* lyr = m_map->operationalLayers()->at(modelIndex);
+
+  // add the layer to the other list
+  m_deletedLayers.append(lyr);
+  m_deletedLayersListNames.append(lyr->name());
+  deletedLayersListChanged();
```
JamesMBallard: This is not too bad.

ArcGISRuntimeSDKQt_QMLSamples/Layers/ManageOperationalLayers/DrawOrderLayerListModel.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+int DrawOrderLayerListModel::mappedIndex(int index) const
+{
+  const QModelIndex sourceIndex = mapToSource(this->index(index, 0));
+  return sourceIndex.row();
+}
+
+bool DrawOrderLayerListModel::lessThan(const QModelIndex& sourceLeft, const QModelIndex& sourceRight) const
+{
+  return sourceLeft.row() > sourceRight.row();
```
JamesMBallard: same question, should this be `>=`?

PR #897, opened 2019-06-10
adding proxy model to qrc

ldanzinger:
@JamesMBallard please merge. I initially missed adding the proxy model to the qrc which was causing the source code view in the viewer to not display these files


JamesMBallard:
Good find.

PR #904, opened 2019-06-27
Jcaccamo/iwa security

DonMoJa:
@ldanzinger @ecmorris Please review.

Thank you!


ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.qml
```diff
@@ -0,0 +1,156 @@
...
+                text: integratedWindowsAuthenticationSampleModel.mapLoadError
+            }
+            Text {
+                id: detailsLabel
+            }
+        }
+        onAccepted: integratedWindowsAuthenticationSampleModel.errorAccepted();
+    }
+
+
```
ecmorris: double newline

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.cpp
```diff
@@ -0,0 +1,173 @@
...
+  emit isLoadingChanged();
+
+  if (forceLogin)
+  {
+    m_portal = new Portal(url, forceLogin, this);
+  }
+  else
+  {
+    m_portal = new Portal(url, this);
+  }
```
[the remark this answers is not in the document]
DonMoJa: It is not in this situation. If 'true' is passed in with www.arcgis.com it will request the user to sign in when it is supposed to be anonymous. I should clarify with a comment.

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.cpp
```diff
@@ -0,0 +1,173 @@
...
+    PortalQueryParametersForItems query;
+    query.setTypes(QList<PortalItemType>() << PortalItemType::WebMap);
+    m_portal->findItems(query);
+  });
+
+  connect(m_portal, &Portal::findItemsCompleted, this, [this](PortalQueryResultSetForItems* results)
+  {
+    m_webmaps = results->itemResults();
+    m_webmapList.clear();
+    for (PortalItem * pI : *m_webmaps)
```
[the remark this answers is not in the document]
DonMoJa: ahhh good catch!

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.cpp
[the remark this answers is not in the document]
DonMoJa: You are most definitely right. Thank you!

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+  m_portal->load();
+}
+
+void IntegratedWindowsAuthentication::loadSelectedWebmap(int index)
+{
+  if (!m_webmaps)
+    return;
+
+  PortalItem* selectedItem = m_webmaps->at(index);
```
ldanzinger: I believe this whole function can be simplified. No need to explicitly load the PortalItem and Maps. Those will load implicitly. I think this function could be as simple as

```
void IntegratedWindowsAuthentication::loadSelectedWebmap(int index)
{
  if (!m_webmaps)
    return;

  if (m_map)
    delete m_map;

  // create map from portal item
  m_map = new Map(m_webmaps->at(index), this);

  // set map on mapview
  m_mapView->setMap(m_map);
}
```

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.qml
```diff
@@ -0,0 +1,156 @@
...
+                background: Rectangle {
+                    color: "white"
+                }
+            }
+
+            Row {
+                Layout.alignment: Qt.AlignHCenter | Qt.AlignVCenter
+                CheckBox {
+                    id: forceLoginBox
+                    checked: false
```
ldanzinger: should try and just use the text property on Checkbox instead of creating a Row with a Checkbox and Text separate

ArcGISRuntimeSDKQt_QMLSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.qml
```diff
@@ -0,0 +1,241 @@
...
+    PortalQueryParametersForItems {
+        id: webmapQuery
+        types: [ Enums.PortalItemTypeWebMap ]
+    }
+
+    function searchPortal (portalUrl, forceLogin) {
+
+        var portal;
+        if (forceLogin) {
+            portal = ArcGISRuntimeEnvironment.createObject("Portal", {url: portalUrl, loginRequired: forceLogin});
```
ldanzinger: same comment as in C++ - you shouldn't need to create a new instance every time. The Portal object isn't really meant to be transient but rather, you would normally have 1 portal instance in your app that you've authenticated with and then you issue your commands on that one instance (query, add, remove, etc). 

I suggest you create the instance once only. Since this is QML, I'd recommend creating it declaratively and connecting to signals there instead of in JS code as it is a little easier to follow and to handle JS scoping/GC issues

ArcGISRuntimeSDKQt_QMLSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.qml
```diff
@@ -0,0 +1,241 @@
...
+
+        });
+
+        portal.findItemsStatusChanged.connect(function() {
+            if ( portal.findItemsStatus === Enums.TaskStatusCompleted ) {
+                indicator.running = false;
+                portalItemListModel = portal.findItemsResult.itemResults;
+                var index = 0
+                var error = portalItemListModel.forEach(function(prtlItem) {
+                    listModelForComboBox[index] = prtlItem.title;
```
ldanzinger: no need to create a `listModelForComboBox` - the itemResults already returns you a list model that should be usable from within a combo box. You can access it via the title role. ComboBox allows you to set the role you want - https://doc.qt.io/archives/qt-5.11/qml-qtquick-controls2-combobox.html#textRole-prop

ArcGISRuntimeSDKQt_QMLSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.qml
```diff
@@ -0,0 +1,241 @@
...
+        });
+
+        portal.load();
+        indicator.running = true;
+    }
+
+    function loadSelectedWebmap(selectedWebmap) {
+        portalItem = selectedWebmap;
+
+        portalItem.loadStatusChanged.connect(createMap);
```
ldanzinger: same comment as C++ -  no need to explicitly load PortalItem or Map. If you construct a Map and give it to the view, the view will kick off the load of the map which will load the portalitem. This should be a pretty simple function:

```
function loadSelectedWebmap(selectedWebmap) {
    mapView.map = ArcGISRuntimeEnvironment.createObject("Map", {"item": selectedWebmap});
}
```

ArcGISRuntimeSDKQt_QMLSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.qml
```diff
@@ -0,0 +1,241 @@
...
+    }*/
+
+    Dialog {
+        id: webMapMsg
+        modal: true
+        x: Math.round(parent.width - width) / 2
+        y: Math.round(parent.height - height) / 2
+        standardButtons: Dialog.Ok
+        title: qsTr("Could not load web map!")
+        property alias text : textLabel.text
```
ldanzinger: property should move up to top under the id

ArcGISRuntimeSDKQt_QMLSamples/CloudAndPortal/IntegratedWindowsAuthentication/README.md
```diff
@@ -0,0 +1,35 @@
+# Integrated windows authentication
```
ldanzinger: Same comments as C++ readme

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.cpp
```diff
@@ -0,0 +1,169 @@
...
+  });
+
+  connect(m_portal, &Portal::findItemsCompleted, this, [this](PortalQueryResultSetForItems* results)
+  {
+    if(!results)
+      return;
+
+    m_webmaps = results->itemResults();
+    m_webmapList.clear();
+    for (PortalItem * pI : *m_webmaps)
```
JamesMBallard: For consistent code style, prefer `PortalItem* pI` to `PortalItem * pI`

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.cpp
```diff
@@ -0,0 +1,173 @@
...
+      emit mapLoadErrorChanged();
+    });
+
+    connect(m_map, &Map::loadStatusChanged, this, [this]()
+    {
+      if (!m_map || m_map->loadStatus() != LoadStatus::Loaded)
+        return;
+
+      m_mapView->setMap(m_map);
+      m_mapView->setVisible(true);
```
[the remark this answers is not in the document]
DonMoJa: Forgot to remove that. Thanks.

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.qml
```diff
@@ -0,0 +1,156 @@
...
+                visible: true
+            }
+
+            Button {
+                text: qsTr("Load Web Map")
+                Layout.fillWidth: true
+                Layout.margins: 2
+                visible: true
+                onClicked: {
+                    if (webmapsList.currentIndex >= 0) {
```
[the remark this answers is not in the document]
DonMoJa: If the user would click "load web map" before searching either the public or secure portal. I have updated it for the button to be disabled unless a model is defined.

DonMoJa:
@ldanzinger code has been updated and is ready for re-review.

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.cpp
```diff
@@ -0,0 +1,209 @@
...
+      emit webmapListModelChanged();
+      emit isLoadingChanged();
+      emit mapLoadErrorChanged();
+      return;
+    }
+
+    m_iwaSecurePortal->findItems(*query);
+  });
+
+  connect(m_iwaSecurePortal, &Portal::findItemsCompleted, this, [this](PortalQueryResultSetForItems* result)
```
ldanzinger: This is pretty much the same code as the public portal's signal handler. Since you need to connect several objects to the same handler, a better option would probably be to either create a SLOT that is called by each, or to take out the logic from each and put it in a separate member function, and call the method from the lambda. 

- Here is how you could create a SLOT and hook it up - https://github.com/Esri/arcgis-runtime-samples-qt/blob/7e53f3d4f69a063cfa08ba092d0732d9f72b0c12/ArcGISRuntimeSDKQt_CppSamples/Scenes/OpenMobileScenePackage/OpenMobileScenePackage.cpp#L164

- alternatively, you could create a member function called `processResults` and call that from here, and stick all this logic in there. That should eliminate the duplication.

Same concept probably applies to the done loading signal handler

ArcGISRuntimeSDKQt_QMLSamples/CloudAndPortal/IntegratedWindowsAuthentication/IntegratedWindowsAuthentication.qml
```diff
@@ -0,0 +1,220 @@
...
+    MapView {
+        id: mapView
+        anchors.fill: parent
+
+        Map {
+            BasemapTopographic {}
+        }
+    }
+
+    Portal {
```
ldanzinger: this looks much better. I think the IWA secured Portal instance should probably be declared as well. Is this not done because you can't re-set the URL in the case that the user provides incorrect URL?

DonMoJa: Precisely.

DonMoJa:
@ldanzinger removed duplicate code where possible with slots. Ready for you to look over again. I created two separate slots for both portals since they both call "findItems()" once the portal is done loading and can't share the same portal.

PR #912, opened 2019-07-31
simplifying load selected webmap function that is similar to one in IWA sample

DonMoJa:
@ldanzinger We had discussed this method a little while back when I was working on the IWA sample. No need to explicitly load the PortalItem and Map as they will load implicitly. Tested the code and it works as it did before.

Please review when you have a moment. Thanks!

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/SearchForWebmap/SearchForWebmap.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/SearchForWebmap/SearchForWebmap.cpp
@@ -141,6 +141,12 @@ void SearchForWebmap::loadSelectedWebmap(int index)
    // create map from portal item
    m_map = new Map(m_webmaps->at(index), this);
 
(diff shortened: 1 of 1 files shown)


ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/SearchForWebmap/SearchForWebmap.cpp
```diff
@@ -133,51 +133,17 @@ void SearchForWebmap::searchNext()
...
-      delete m_map;
-      m_map = nullptr;
-    }
-
-    m_mapLoadeError.clear();
-    emit mapLoadErrorChanged();
-
-    m_map = new Map(m_selectedItem, this);
-
-    connect(m_map, &Map::errorOccurred, this, [this]()
```
ldanzinger: do we still want to have this in the sample to catch map errors?

DonMoJa: Good point. I will put that back in.

PR #921, opened 2019-08-21
Ldanzinger/mmpk expiration

ldanzinger:
@DonMoJa please review. 90% is copied from the open mobile map package sample - the unique bits are checking the MobileMapPackage::expiration info and displaying in the UI


ArcGISRuntimeSDKQt_CppSamples/Maps/HonorMobileMapPackageExpiration/HonorMobileMapPackageExpiration.cpp
```diff
@@ -0,0 +1,170 @@
...
+      return;
+    }
+
+    if (!m_mobileMapPackage || !m_mapView || m_mobileMapPackage->maps().isEmpty())
+    {
+      return;
+    }
+
+    // Access Expiration Info
+    Expiration expiration = m_mobileMapPackage->expiration();
```
DonMoJa: Should we test to see if the returned expiration is empty before proceeding to insure it is valid? 

if (expiration.isEmpty())
  // return and do nothing?

ldanzinger: good call. added that in to the latest commit

PR #922, opened 2019-08-23
Fix missing commas in metadata json

noahmulfinger:
Minor issue in metadata file for the LoadWfsXmlQuery sample. Noticed when running the developers website.


ldanzinger:
Good catch. Looks good to merge.

PR #927, opened 2019-09-06
Jcaccamo/save kml screenshot fix

DonMoJa:
@JamesMBallard please review when you have a moment. Updated the screenshot and realized that a dialog confirming the file was saved would be useful information to convey.


ArcGISRuntimeSDKQt_QMLSamples/Layers/CreateAndSaveKmlFile/CreateAndSaveKmlFile.qml
```diff
@@ -57,6 +57,10 @@ Rectangle {
                 if (saveStatus === Enums.TaskStatusErrored) {
                     console.log(`Error: ${error.message} - ${error.additionalMessage}`);
                 }
+
+                if (saveStatus === Enums.TaskStatusCompleted) {
```
JamesMBallard: This can be an `else if` statement. These are mutually exclusive states.

Or, the `Enums.TaskStatusErrored` case can return.

PR #942, opened 2019-11-13
Jcaccamo/preplanned map

DonMoJa:
@ldanzinger @lsmallwood Please review when you have some time. Thank you!


ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.h
```diff
@@ -0,0 +1,92 @@
...
+class OfflineMapTask;
+class PreplannedMapArea;
+class DownloadPreplannedOfflineMapJob;
+class MobileMapPackage;
+class SimpleLineSymbol;
+}
+}
+
+#include <QObject>
+#include <QTemporaryDir>
```
lsmallwood: we normally alphabetize includes and separate Qt and API headers with a blank line

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.h
```diff
@@ -0,0 +1,92 @@
...
+  Q_PROPERTY(bool busy MEMBER m_busy NOTIFY busyChanged)
+  Q_PROPERTY(bool mapExists MEMBER m_mapExists NOTIFY mapExistsChanged)
+  Q_PROPERTY(bool viewingOnlineMaps MEMBER m_viewingOnlineMaps NOTIFY viewingOnlineMapsChanged())
+
+public:
+  explicit DownloadPreplannedMap(QObject* parent = nullptr);
+  ~DownloadPreplannedMap();
+
+  static void init();
+  Q_INVOKABLE void downloadMapArea(const int index);
```
lsmallwood: I don't think we need to make these `int` arguments const

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+// limitations under the License.
+// [Legal]
+
+#include "DownloadPreplannedMap.h"
+
+#include "Map.h"
+#include "MapQuickView.h"
+#include "OfflineMapTask.h"
+#include "PreplannedMapArea.h"
+#include "PreplannedMapAreaListModel.h"
```
lsmallwood: alphabetize

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+DownloadPreplannedMap::DownloadPreplannedMap(QObject* parent /* = nullptr */):
+  QObject(parent),
+  m_portalItem(new PortalItem("acc027394bc84c2fb04d1ed317aac674", this)),
+  m_lineSymbol(new SimpleLineSymbol(SimpleLineSymbolStyle::Solid, QColor(Qt::red), 5, this)),
+  m_graphicParent(new QObject())
+{
+  m_busy = true;
+  emit busyChanged();
+
+  m_graphicsOverlay = new GraphicsOverlay(this);
```
lsmallwood: you could add `m_busy` and `m_graphicsOverlay` to the initializer list

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+  if (!mapView || mapView == m_mapView)
+    return;
+
+  m_mapView = mapView;
+  m_mapView->setMap(m_map);
+
+  emit mapViewChanged();
+}
+
+void DownloadPreplannedMap::downloadMapArea(int index)
```
lsmallwood: I think you have this as `const int index` in the header. I think it's fine to have this as non-const

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+{
+  if (!m_offlineMapTask)
+    return;
+
+  if (m_offlineMapTask->preplannedMapAreaList()->isEmpty())
+    return;
+
+  if (m_offlineMapTask->preplannedMapAreaList()->at(index)->loadStatus() != LoadStatus::Loaded)
```
lsmallwood: Probably better to get the area so you don't have to keep fetching it from the  model:

```suggestion
  PreplannedMapArea* mapArea = m_offlineMapTask->preplannedMapAreaList()->at(index);
  if (!mapArea || mapArea->loadStatus() != LoadStatus::Loaded)
```

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+    return;
+
+  m_busy = true;
+  emit busyChanged();
+
+  QString path = QString(m_tempPath.path() + "/" + m_offlineMapTask->preplannedMapAreaList()->at(index)->portalItem()->title() + ".mmpk");
+
+  QDir tempDir(path);
+
+  if (m_mapExists)
```
lsmallwood: I don't understand this? Is the idea that, if you've already downloaded the area we just display it?

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+      m_params = DownloadPreplannedOfflineMapParameters(parameters);
+      m_params.setUpdateMode(PreplannedUpdateMode::NoUpdates);
+
+      m_preplannedMapJob = m_offlineMapTask->downloadPreplannedOfflineMap(m_params, path);
+      connect(m_preplannedMapJob, &DownloadPreplannedOfflineMapJob::jobDone, this, &DownloadPreplannedMap::onDownloadPreplannedMapJobCompleted);
+      m_preplannedMapJob->start();
+    });
+
+    m_offlineMapTask->createDefaultDownloadPreplannedOfflineMapParameters(m_offlineMapTask->preplannedMapAreaList()->at(index));
```
lsmallwood: use the cached area again here

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+    });
+
+    m_mmpk->load();
+    return;
+  }
+  else
+  {
+    connect(m_offlineMapTask, &OfflineMapTask::createDefaultDownloadPreplannedOfflineMapParametersCompleted, this, [this, path] (QUuid ,const DownloadPreplannedOfflineMapParameters &parameters)
+    {
+
```
lsmallwood: remove empty line

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+      m_viewingOnlineMaps = false;
+      emit viewingOnlineMapsChanged();
+    });
+
+    m_mmpk->load();
+    return;
+  }
+  else
+  {
+    connect(m_offlineMapTask, &OfflineMapTask::createDefaultDownloadPreplannedOfflineMapParametersCompleted, this, [this, path] (QUuid ,const DownloadPreplannedOfflineMapParameters &parameters)
```
lsmallwood:
```suggestion
    connect(m_offlineMapTask, &OfflineMapTask::createDefaultDownloadPreplannedOfflineMapParametersCompleted, this, [this, path] (QUuid ,const DownloadPreplannedOfflineMapParameters& parameters)
```

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+    {
+
+      m_params = DownloadPreplannedOfflineMapParameters(parameters);
```
lsmallwood: I think you should just be able to do:
```suggestion
      m_params = parameters;
```

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+      m_params.setUpdateMode(PreplannedUpdateMode::NoUpdates);
```
lsmallwood: Let's add a comment here to explain that we are downloading without registering updates for the purposes of the sample - but also explain the other options

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+      m_preplannedMapJob = m_offlineMapTask->downloadPreplannedOfflineMap(m_params, path);
+      connect(m_preplannedMapJob, &DownloadPreplannedOfflineMapJob::jobDone, this, &DownloadPreplannedMap::onDownloadPreplannedMapJobCompleted);
+      m_preplannedMapJob->start();
+    });
+
+    m_offlineMapTask->createDefaultDownloadPreplannedOfflineMapParameters(m_offlineMapTask->preplannedMapAreaList()->at(index));
+  }
+}
+
+void DownloadPreplannedMap::checkIfMapExists(const int index)
```
lsmallwood:
```suggestion
void DownloadPreplannedMap::checkIfMapExists(int index)
```

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+  if (!m_offlineMapTask)
+    return;
+
+  if (m_offlineMapTask->preplannedMapAreaList()->isEmpty())
+    return;
+
+  if (m_offlineMapTask->preplannedMapAreaList()->at(index)->loadStatus() != LoadStatus::Loaded)
+    return;
+
+  QString path = QString(m_tempPath.path() + "/" + m_offlineMapTask->preplannedMapAreaList()->at(index)->portalItem()->title() + ".mmpk");
```
lsmallwood: this looks to be duplicating the logic for the download path from above? I think it would be better to refactor this out so we can share that logic

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+void DownloadPreplannedMap::checkIfMapExists(const int index)
+{
+  if (!m_offlineMapTask)
+    return;
+
+  if (m_offlineMapTask->preplannedMapAreaList()->isEmpty())
+    return;
+
+  if (m_offlineMapTask->preplannedMapAreaList()->at(index)->loadStatus() != LoadStatus::Loaded)
```
lsmallwood: create a variable for the mapArea again

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+    return;
+
+  QString path = QString(m_tempPath.path() + "/" + m_offlineMapTask->preplannedMapAreaList()->at(index)->portalItem()->title() + ".mmpk");
+  QDir tempDir(path);
+
+  if (tempDir.exists())
```
lsmallwood: I think you could simplify this logic to something like:
```suggestion
m_mapExists = tempDir.exists();
emit mapExistsChanged();
```

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+  {
+    m_mapExists = false;
+    emit mapExistsChanged();
+  }
+
+  if (m_viewingOnlineMaps)
+    m_mapView->setViewpointGeometry(m_offlineMapTask->preplannedMapAreaList()->at(index)->areaOfInterest());
+}
+
+void DownloadPreplannedMap::showOnlineMap(const int index)
```
lsmallwood:
```suggestion
void DownloadPreplannedMap::showOnlineMap(int index)
```

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+{
+  m_viewingOnlineMaps = true;
+  emit viewingOnlineMapsChanged();
+  m_mapView->setMap(m_map);
+  m_offlineMapTask->preplannedMapAreaList()->at(index)->areaOfInterest();
```
lsmallwood: not sure what this line is doing? Should it be:

```suggestion
  Geometry areaOfInterest = m_offlineMapTask->preplannedMapAreaList()->at(index)->areaOfInterest();
```

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+  {
+    connect(mapArea, &PreplannedMapArea::doneLoading, this, [this, mapArea] (Error e)
+    {
+      if (!e.isEmpty())
+      {
+        qDebug() << e.message() << " - " << e.additionalMessage();
+        return;
+      }
+
+      Graphic* areaOfInterest = new Graphic(mapArea->areaOfInterest(), m_graphicParent.get());
```
lsmallwood: I don't see anywhere that we actually delete the `m_graphicParent` so might be simpler just to remove that and use `this` instead?

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,111 @@
...
+
+Item {
+
+    // add a mapView component
+    MapView {
+        id: view
+        anchors.fill: parent
+    }
+
+    // Declare the C++ instance which creates the scene etc. and supply the view
```
lsmallwood:
```suggestion
    // Declare the C++ instance which creates the map etc. and supply the view
```

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## About the data
+
+The [Naperville stormwater network map](https://arcgisruntime.maps.arcgis.com/home/item.html?id=acc027394bc84c2fb04d1ed317aac674) is based on ArcGIS Solutions for Stormwater and provides a realistic depiction of a theoretical stormwater network.
+
+## Additional information
+
+`PreplannedUpdateMode` can be used to set the way the preplanned map area receives updates in several ways:
+
+* `NO_UPDATES` - No updates will be performed.
```
lsmallwood: In Qt API these would not be capitalized or separated with underscores:
```suggestion
* `NoUpdates` - No updates will be performed.
```

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+    MapView {
+        id: mapView
+        anchors.fill: parent
+
+        Map {
+            id: onlineMap
+            item: portalItem
+        }
+
+        Portal {
```
lsmallwood: I don't think you need to declare a `Portal` if it's the default `arcgis.com`. You could also move the portal and the item outside of the MapView scope

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+
+                preplannedCombo.model = offlineMapTask.preplannedMapAreaList;
+                busy.visible = false;
+
+                for (let i = 0; i < offlineMapTask.preplannedMapAreaList.count; i++) {
+                    offlineMapTask.preplannedMapAreaList.get(i).loadStatusChanged.connect(function () {
+                        if (offlineMapTask.preplannedMapAreaList.get(i).loadStatus !== Enums.LoadStatusLoaded)
+                            return;
+
+                        if (i === 0)
```
lsmallwood: why only for index 0? Wouldn't you want to do this for any index the user supplies?

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+
+            onPreplannedMapAreasStatusChanged: {
+                if (preplannedMapAreasStatus !== Enums.TaskStatusCompleted)
+                    return;
+
+                preplannedCombo.model = offlineMapTask.preplannedMapAreaList;
+                busy.visible = false;
+
+                for (let i = 0; i < offlineMapTask.preplannedMapAreaList.count; i++) {
+                    offlineMapTask.preplannedMapAreaList.get(i).loadStatusChanged.connect(function () {
```
lsmallwood: I think you can use a local variable to store the mapArea so you don't have to keep grabbing it from the model:
```suggestion
                    let mapArea = offlineMapTask.preplannedMapAreaList.get(i);
                    mapArea.loadStatusChanged.connect(function () {
```

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+                    });
+                    offlineMapTask.preplannedMapAreaList.get(i).load();
+                }
+            }
+
+            onCreateDefaultDownloadPreplannedOfflineMapParametersStatusChanged: {
+                if (createDefaultDownloadPreplannedOfflineMapParametersStatus !== Enums.TaskStatusCompleted)
+                    return;
+
+                createDefaultDownloadPreplannedOfflineMapParametersResult.updateMode = Enums.PreplannedUpdateModeNoUpdates;
```
lsmallwood: add comment explaining updateMode

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+                let result = createDefaultDownloadPreplannedOfflineMapParametersResult;
+
+                path = outputMapPackage +"/" + result.preplannedMapArea.portalItem.title + ".mmpk";
```
lsmallwood: this should not get an `.mmpk` extension because it will download to a directory rather than an archived file. We have the same problem in our current `GenerateOfflineMap` sample I think. Might be good to clean that up at the same time?

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+  if (m_offlineMapTask->preplannedMapAreaList()->isEmpty())
+    return;
+
+  if (m_offlineMapTask->preplannedMapAreaList()->at(index)->loadStatus() != LoadStatus::Loaded)
+    return;
+
+  m_busy = true;
+  emit busyChanged();
+
+  QString path = QString(m_tempPath.path() + "/" + m_offlineMapTask->preplannedMapAreaList()->at(index)->portalItem()->title() + ".mmpk");
```
lsmallwood: we shouldn't append the `.mmpk` suffix here because the OfflineMapTask downloads to a directory (e.g. an unpacked mobileMapPackage) rather than a directory.

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+                            return;
+
+                        if (mmpk.maps.length < 1)
+                            return;
+
+                        mapView.map = mmpk.maps[0];
+                        busy.visible = false;
+                    });
+                    mmpk.load();
+                    return;
```
lsmallwood: don't need the `return`

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+                    mapExists = true;
+                    busy.visible = false;
+                });
+
+                job.start();
+            }
+        }
+
+        Component.onCompleted: {
+            portalItem.load();
```
lsmallwood: I think the portalItem should get loaded by the map once it is set to the mapView

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+                text: mapExists ? qsTr("View Preplanned area") : qsTr("Download Preplanned Area")
+                onClicked: {
+                    preplannedArea = offlineMapTask.preplannedMapAreaList.get(preplannedCombo.currentIndex);
+                    offlineMapTask.createDefaultDownloadPreplannedOfflineMapParameters(preplannedArea);
+                    graphicsOverlay.visible = false;
+                    viewingOnlineMaps = false;
+                    busy.visible = true;
+                }
+
+                onWidthChanged: {
```
lsmallwood: can this be removed?

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## About the data
+
+The [Naperville stormwater network map](https://arcgisruntime.maps.arcgis.com/home/item.html?id=acc027394bc84c2fb04d1ed317aac674) is based on ArcGIS Solutions for Stormwater and provides a realistic depiction of a theoretical stormwater network.
+
+## Additional information
+
+`PreplannedUpdateMode` can be used to set the way the preplanned map area receives updates in several ways:
+
+* `NO_UPDATES` - No updates will be performed.
```
lsmallwood: enum names are wrong for QML

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+                            return;
+
+                        mapView.map = mmpk.maps[0];
+                        busy.visible = false;
+                    });
+                    mmpk.load();
+                    return;
+                }
+
+                var job = offlineMapTask.downloadPreplannedOfflineMapWithParameters(createDefaultDownloadPreplannedOfflineMapParametersResult, path);
```
ldanzinger: you might want to see about storing this variable at global scope. It's possible the function returns and the gc runs before the jobStatusChanged.connect function executes

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+        ColumnLayout {
+            Button {
+                id: onlineMapButton
+                Layout.fillWidth: true
+                Layout.margins: 2
+                Layout.alignment: Qt.AlignHCenter
+                text: qsTr("Show Online Map")
+                enabled: !busy.visible & !viewingOnlineMaps
+                onClicked: {
+
```
ldanzinger: extra space can be removed

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+            Button {
+                id: downloadOrView
+                Layout.fillWidth: true
+                Layout.margins: 2
+                enabled: !busy.visible
+                text: mapExists ? qsTr("View Preplanned area") : qsTr("Download Preplanned Area")
+                onClicked: {
+                    preplannedArea = offlineMapTask.preplannedMapAreaList.get(preplannedCombo.currentIndex);
+                    offlineMapTask.createDefaultDownloadPreplannedOfflineMapParameters(preplannedArea);
```
ldanzinger: I think more code comments could be added throughout. There are quite a few tasks kicked off and they are a bit buried around other app logic code, so it is hard to pick out the important code

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    BusyIndicator {
+        id: busy
+        anchors.centerIn: parent
+        visible: false;
+    }
+
+    function checkIfFileExists(index) {
+        path = outputMapPackage + "/" + offlineMapTask.preplannedMapAreaList.get(index).portalItem.title + ".mmpk";
```
ldanzinger: Same comment from Luke about the .mmpk suffix probably applies here

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+        fileFolder.url = path;
+        mapExists = fileFolder.exists;
```
ldanzinger: the function name seems like it would return a bool, but it seems like it returns void and sets some variable bool. Maybe this can be cleaned up

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    FileFolder {
+        id: fileFolder
+    }
+
+    BusyIndicator {
+        id: busy
+        anchors.centerIn: parent
+        visible: false;
```
ldanzinger:
```suggestion
        visible: false
```

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    function checkIfFileExists(index) {
+        path = outputMapPackage + "/" + offlineMapTask.preplannedMapAreaList.get(index).portalItem.title + ".mmpk";
+        fileFolder.url = path;
+        mapExists = fileFolder.exists;
+        return;
```
ldanzinger: return not needed, unless you decide to return a bool

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
DonMoJa: Mistake on my part. Didn't need to be in there but I'll make the suggested changes.

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,111 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// [Legal]
+
+import QtQuick 2.6
+import QtQuick.Controls 2.12
+import QtQuick.Layouts 1.12
+import Esri.Samples 1.0
+import Esri.ArcGISExtras 1.1
```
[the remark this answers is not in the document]
DonMoJa: artifact from copying some of the other imports from the xml sample. Thanks for catching that!

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,111 @@
...
+            Text {
+                text: qsTr("Download(s) deleted on exit")
+                color: "white"
+                Layout.alignment: Qt.AlignHCenter
+                Layout.margins: 2
+            }
+        }
+    }
+
+    FileFolder {
```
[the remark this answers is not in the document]
DonMoJa: Ah yes it can!

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
DonMoJa: yeah I bounced between two ways of doing this and ended up having a little of both I see.

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
DonMoJa: Yes precisely. The signature of the method could be improved to be a little less specific or better yet, have the logic broken down into two separate methods specific to their task.

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.pro
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#-------------------------------------------------
+
+TEMPLATE = app
+
+# additional modules are pulled in via arcgisruntime.pri
+QT += opengl qml quick
+
+CONFIG += c++11
```
JamesMBallard: For 100.7:

```suggestion
CONFIG += c++14
```

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.pro
```diff
@@ -0,0 +1,64 @@
...
+# limitations under the License.
+#-------------------------------------------------
+
+mac {
+    cache()
+}
+
+#-------------------------------------------------------------------------------
+
+CONFIG += c++11
```
JamesMBallard:
```suggestion
CONFIG += c++14
```

DonMoJa:
> 100.7 can now use the C++14 standard.

Ah yes! I need to go update my template so I don't keep running into this.

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
DonMoJa: Ah yes this. I can remove this now. Originally I was doing this to make sure that the default selected item in the combo box would be checked to see if it already existed( They used to not be deleted upon exit). Now this is not necessary. Thanks!

DonMoJa:
@ldanzinger @lsmallwood made requested changes. Ready for re-review. Thanks!

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -84,103 +86,67 @@ void DownloadPreplannedMap::setMapView(MapQuickView* mapView)
...
-  if (m_mapExists)
-  {
-    if (m_mmpk)
-    {
-      delete m_mmpk;
-      m_mmpk = nullptr;
-    }
-
-    m_mmpk = new MobileMapPackage(path, this);
+  loadSelectedMap(*mapArea, index);
```
lsmallwood: I think this should just take the mapArea as a ptr rather than dereferencing

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -84,103 +86,67 @@ void DownloadPreplannedMap::setMapView(MapQuickView* mapView)
...
+}
 
-    connect(m_mmpk, &MobileMapPackage::doneLoading, this, [this] (Error e)
-    {
-      if (!e.isEmpty())
-        qDebug() << e.message() << " - " << e.additionalMessage();
+void DownloadPreplannedMap::loadSelectedMap(const PreplannedMapArea& mapArea, int index)
```
lsmallwood: `void DownloadPreplannedMap::loadSelectedMap(PreplannedMapArea* mapArea, int index)`

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -217,10 +183,61 @@ void DownloadPreplannedMap::loadPreplannedMapAreas()
...
       m_graphicsOverlay->graphics()->append(areaOfInterest);
     });
 
     mapArea->load();
   }
 }
+
+void DownloadPreplannedMap::downloadPreplannedMapArea(const QString& path, int index)
+{
+  connect(m_offlineMapTask, &OfflineMapTask::createDefaultDownloadPreplannedOfflineMapParametersCompleted, this, [this, path] (QUuid ,const DownloadPreplannedOfflineMapParameters& parameters)
```
lsmallwood: if this method get's called multiple times, you will end up with multiple connections for this. Might be better to move the connection to the point where you first create the `m_offlineMapTask` so that it will be setup once only

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -217,10 +183,61 @@ void DownloadPreplannedMap::loadPreplannedMapAreas()
...
+
+    /* Set the update mode to not receive updates.
+     * Other options:
+     * SyncWithFeatureServices - changes will be synced directly with the
+     * underlying feature services.
+     * DownloadScheduledUpdates - schedulded, read-only updates will be
+     * downloaded and applied to the local geodatabase. */
+    m_params.setUpdateMode(PreplannedUpdateMode::NoUpdates);
+
+    m_preplannedMapJob = m_offlineMapTask->downloadPreplannedOfflineMap(m_params, path);
```
lsmallwood: I don't see anywhere that you are handling progress from the job. That's fine but I think the README suggests that progress will be displayed?

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.h
```diff
@@ -72,6 +73,10 @@ private slots:
   Esri::ArcGISRuntime::MapQuickView* mapView() const;
   void setMapView(Esri::ArcGISRuntime::MapQuickView* mapView);
 
+  void loadSelectedMap(const Esri::ArcGISRuntime::PreplannedMapArea& mapArea, int index);
```
lsmallwood: void loadSelectedMap(PreplannedMapArea* mapArea, int index);

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -136,31 +123,43 @@ Rectangle {
...
+                    // Start job to take preplanned map area offline.
+                    job.start();
+                }
             }
         }
+    }
+
+    PortalItem {
+        id: portalItem
+        portal: portal
```
lsmallwood: I'm not sure you have `portal` declared anymore?

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,243 @@
...
+  emit busyChanged();
+
+  loadSelectedMap(*mapArea, index);
+}
+
+void DownloadPreplannedMap::loadSelectedMap(const PreplannedMapArea& mapArea, int index)
+{
+  const QString path = QString(m_tempPath.path() + "/" + mapArea.portalItem()->title());
+
+  QDir tempDir(path);
```
ldanzinger: const

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+      // fetch preplanned map areas from the portal item
+      m_offlineMapTask->preplannedMapAreas();
+    });
+
+    // Load offline map task.
+    m_offlineMapTask->load();
+  });
+
+  m_map = new Map(m_portalItem, this);
```
ldanzinger: can be moved to initializer list

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,226 @@
...
+    return;
+
+  if (m_offlineMapTask->preplannedMapAreaList()->isEmpty())
+    return;
+
+  if (m_offlineMapTask->preplannedMapAreaList()->at(index)->loadStatus() != LoadStatus::Loaded)
+    return;
+
+  QString path = QString(m_tempPath.path() + "/" + m_offlineMapTask->preplannedMapAreaList()->at(index)->portalItem()->title() + ".mmpk");
+  QDir tempDir(path);
```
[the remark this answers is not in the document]
ldanzinger: is this resolved? I think tempDir can still be const

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
```diff
@@ -0,0 +1,243 @@
...
+}
+
+void DownloadPreplannedMap::loadPreplannedMapAreas()
+{
+  m_preplannedList = m_offlineMapTask->preplannedMapAreaList();
+  m_busy = false;
+  emit preplannedListChanged();
+  emit busyChanged();
+
+  for (PreplannedMapArea* mapArea : *static_cast<PreplannedMapAreaListModel*>(m_preplannedList))
```
ldanzinger: Probably is better to pre-cache `*static_cast<PreplannedMapAreaListModel*>(m_preplannedList)` before the loop

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -0,0 +1,279 @@
...
+                 * DownloadScheduledUpdates - schedulded, read-only updates will be
+                 * downloaded and applied to the local geodatabase. */
+                createDefaultDownloadPreplannedOfflineMapParametersResult.updateMode = Enums.PreplannedUpdateModeNoUpdates;
+                let result = createDefaultDownloadPreplannedOfflineMapParametersResult;
+
+                path = outputMapPackage +"/" + result.preplannedMapArea.portalItem.title;
+                fileFolder.url = path;
+
+                if (fileFolder.exists) {
+                    var mmpk = ArcGISRuntimeEnvironment.createObject("MobileMapPackage", { path: path });
```
ldanzinger: might be good to cache this as a global variable. Seems like you'd want the MMPK to persist throughout the lifetime of the sample?

ArcGISRuntimeSDKQt_CppSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.cpp
DonMoJa: Ah good catch! I can add that in easily.

DonMoJa:
@ldanzinger @lsmallwood files updates with the requested changes.

PR #956, opened 2019-12-20
using new mapview callout data property directly

ldanzinger:
@JamesMBallard please review/merge once API addition is merged. FYI - I will update the guide doc that uses the snippet that is removed here since it will no longer be required

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/EditData/DeleteFeaturesFeatureService/DeleteFeaturesFeatureService.h
+++ b/ArcGISRuntimeSDKQt_CppSamples/EditData/DeleteFeaturesFeatureService/DeleteFeaturesFeatureService.h
@@ -20,8 +20,7 @@
 namespace Esri
 {
   namespace ArcGISRuntime
-  {
-    class CalloutData;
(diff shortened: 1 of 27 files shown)


ArcGISRuntimeSDKQt_CppSamples/EditData/DeleteFeaturesFeatureService/DeleteFeaturesFeatureService.h
```diff
@@ -49,13 +46,11 @@ class DeleteFeaturesFeatureService : public QQuickItem
   Q_INVOKABLE void deleteSelectedFeature();
 
 signals:
-  void calloutDataChanged();
```
JamesMBallard: This still works correctly with this?

ldanzinger: Yes. Tested all of them. The CalloutData pointer itself doesn't change but the underlying properties in the CalloutData do, and those all emit

PR #965, opened 2020-02-04
Jcaccamo/subtype fl

jared-esri:
@ldanzinger @JamesMBallard Please review when you have a moment.

Not 100% sold on the slot I added for once the feature layer was done loading. Appreciate any thoughts or insights you might have.


ArcGISRuntimeSDKQt_QMLSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.qml
```diff
@@ -0,0 +1,192 @@
...
+Rectangle {
+    id: rootRectangle
+    clip: true
+    width: 800
+    height: 600
+
+    readonly property var labelJson: { "labelExpression": "[nominalvoltage]", "labelPlacement": "esriServerPointLabelPlacementAboveRight", "useCodedValues": true, "symbol": { "angle": 0, "backgroundColor": [ 0, 0, 0, 0 ], "borderLineColor": [ 0, 0, 0, 0 ], "borderLineSize": 0, "color": [ 0, 0, 255, 255 ], "font": { "decoration": "none", "size": 10.5, "style": "normal", "weight": "normal" }, "haloColor": [ 255, 255, 255, 255 ], "haloSize": 2, "horizontalAlignment": "center", "kerning": false, "type": "esriTS", "verticalAlignment": "middle", "xoffset": 0, "yoffset": 0 } }
+    property var subtypeSublayer
+    property var originalRenderer
+    property var mapScale: mapView ? Math.round(mapView.mapScale) : 0
```
JamesMBallard:
```suggestion
    property double mapScale: mapView ? Math.round(mapView.mapScale) : 0
```

https://doc.qt.io/qt-5/qml-double.html

ArcGISRuntimeSDKQt_QMLSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.qml
```diff
@@ -0,0 +1,192 @@
...
+    property var sublayerMinScale
```
JamesMBallard:
```suggestion
    property double sublayerMinScale
```

ArcGISRuntimeSDKQt_QMLSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.qml
```diff
@@ -0,0 +1,192 @@
...
+                    var labelDefinition = ArcGISRuntimeEnvironment.createObject("LabelDefinition", { json : labelJson});
+
+                    if (!labelDefinition || !subtypeSublayer)
+                      return;
+
+                    subtypeSublayer.labelDefinitions.append(labelDefinition);
+                    subtypeSublayer.labelsEnabled = true;
+
+                    // get the original renderer of the sublayer
+                    originalRenderer = subtypeSublayer.renderer;
```
JamesMBallard: I would think since this is not needed in Qml, it is not needed for C++, either.

ArcGISRuntimeSDKQt_CppSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.h
```diff
@@ -0,0 +1,85 @@
...
+  Esri::ArcGISRuntime::SubtypeFeatureLayer* m_subtypeFeatureLayer = nullptr;
+  Esri::ArcGISRuntime::SubtypeSublayer* m_subtypeSublayer = nullptr;
+  Esri::ArcGISRuntime::LabelDefinition* m_labelDefinition = nullptr;
+  Esri::ArcGISRuntime::Renderer* m_originalRenderer = nullptr;
+  Esri::ArcGISRuntime::SimpleRenderer* m_alternateRenderer = nullptr;
+  const QString m_labelJson = QStringLiteral("{ \"labelExpression\":\"[nominalvoltage]\",\"labelPlacement\":\"esriServerPointLabelPlacementAboveRight\",\"useCodedValues\":true,\"symbol\":{\"angle\":0,\"backgroundColor\":[0,0,0,0],\"borderLineColor\":[0,0,0,0],\"borderLineSize\":0,\"color\":[0,0,255,255],\"font\":{\"decoration\":\"none\",\"size\":10.5,\"style\":\"normal\",\"weight\":\"normal\"},\"haloColor\":[255,255,255,255],\"haloSize\":2,\"horizontalAlignment\":\"center\",\"kerning\":false,\"type\":\"esriTS\",\"verticalAlignment\":\"middle\",\"xoffset\":0,\"yoffset\":0}}");
+
+  double m_mapScale = 0.0;
+  double m_sublayerMinScale = 0.0;
+  bool m_busy = true;
```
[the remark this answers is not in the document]
jared-esri: Good point. It was intentional but I will just adjust the logic so it works the other way around. I think it will be more clear that way.

ArcGISRuntimeSDKQt_CppSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.cpp
```diff
@@ -0,0 +1,147 @@
+// [WriteFile Name=DisplaySubtypeFeatureLayer, Category=Layers]
+// [Legal]
+// Copyright 2019 Esri.
```
ldanzinger: Copyrights for new files should be 2020

ArcGISRuntimeSDKQt_CppSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.cpp
```diff
@@ -0,0 +1,143 @@
...
+  m_labelDefinition = LabelDefinition::fromJson(m_labelJson, this);
+
+  if (!m_labelDefinition || !m_subtypeSublayer)
+    return;
+
+  m_subtypeSublayer->labelDefinitions()->append(m_labelDefinition);
+  m_subtypeSublayer->setLabelsEnabled(true);
+
+  // get the original renderer of the sublayer
+  m_originalRenderer = m_subtypeSublayer->renderer()->clone(this);
```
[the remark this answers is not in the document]
jared-esri: In what situation should clone be used?

ArcGISRuntimeSDKQt_CppSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.cpp
```diff
@@ -0,0 +1,147 @@
...
+  // create the feature layer using the feature table
+  m_subtypeFeatureLayer = new SubtypeFeatureLayer(featureTable, this);
+
+  // add the feature layer to the map
+  m_map->operationalLayers()->append(m_subtypeFeatureLayer);
+
+  // set the viewpoint to Naperville, Illinois
+  m_map->setInitialViewpoint(Viewpoint(Envelope(-9812691.11079696, 5128687.20710657, -9812377.9447607, 5128865.36767282, SpatialReference::webMercator())));
+
+  // when subtype feature layer is loaded get the subtype sublayer street lights and define it's labels
```
ldanzinger: confusing, but `it's` is only for contraction, not for possession

```suggestion
  // when subtype feature layer is loaded get the subtype sublayer street lights and define its labels
```

ArcGISRuntimeSDKQt_CppSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.cpp
```diff
@@ -0,0 +1,147 @@
...
+  connect(m_subtypeFeatureLayer, &SubtypeFeatureLayer::doneLoading, this, &DisplaySubtypeFeatureLayer::getSubtypeSublayerAndDefineLabels);
+
+  connect(m_map, &Map::doneLoading, this, [this](Error e)
+  {
+    if (!e.isEmpty())
+      return;
+
+    connect(m_mapView, &MapQuickView::mapScaleChanged, this, [this]()
```
ldanzinger: is there a reason you can only make this connection inside of Map::doneLoading?

ArcGISRuntimeSDKQt_CppSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.cpp
```diff
@@ -0,0 +1,147 @@
...
+void DisplaySubtypeFeatureLayer::getSubtypeSublayerAndDefineLabels(const Error& e)
+{
+  if (!e.isEmpty())
+    return;
+
+  m_busy = false;
+  emit busyChanged();
+
+  // get the Street Light sublayer and define its labels
+  m_subtypeSublayer = m_subtypeFeatureLayer->sublayerWithSubtypeName("Street Light", this);
```
ldanzinger: is there any scenario `m_subtypeFeatureLayer` could be null here?

ArcGISRuntimeSDKQt_CppSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.qml
```diff
@@ -0,0 +1,132 @@
...
+                    Layout.margins: 2
+                    Layout.alignment: Qt.AlignLeft
+                    enabled: !busy.visible
+                    onCheckedChanged: {
+                        if (checked)
+                            model.setAlternativeRenderer();
+                    }
+                }
+
+                Shape {
```
ldanzinger: nice 👍

ArcGISRuntimeSDKQt_QMLSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.qml
jared-esri: Not sure what you mean here. We need the original renderer to be preserved because you can switch between the default renderer and the alternate renderer in the sample. Once you apply the alternate renderer we will lose the default and won't be able to switch back. Is my understanding of this correct?

ArcGISRuntimeSDKQt_QMLSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.qml
```diff
@@ -0,0 +1,192 @@
...
+                    text: qsTr("Show sublayer")
+                    Layout.margins: 2
+                    Layout.alignment: Qt.AlignLeft
+                    checked: true
+                    enabled: subtypeFeatureLayer.loadStatus === Enums.LoadStatusLoaded ? true : false
+                    onCheckedChanged: subtypeSublayer.visible = !subtypeSublayer.visible
+                }
+
+                RadioButton {
+                    text: qsTr("Show original rednerer")
```
ldanzinger: typo here
```suggestion
                    text: qsTr("Show original renderer")
```

ArcGISRuntimeSDKQt_CppSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.cpp
jared-esri: Yes. if the name passed in as a parameter is not found it will return null i believe. That's why I have added checks for the renderer as well as setting the min scale.

ArcGISRuntimeSDKQt_CppSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.cpp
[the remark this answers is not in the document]
JamesMBallard: Clone would be useful if you want to use the same renderer in two different places at the same time. That won't work so you would need to clone it. In your case, you are not doing that.

Try the logic without the clone and see if it works.

ArcGISRuntimeSDKQt_QMLSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.qml
JamesMBallard: Oops, poorly written comment.

I was referring to the clone method used in C++. Since clone is not needed here, that should be an argument that clone is not needed in C++, either.

jared-esri: Ah I see what you mean now. yeah it works just fine without the clone. Thank you for the added clarification on when clone should be used!

jared-esri:
I think all requested changes have been made.

@ldanzinger @JamesMBallard If you could pass over this one more time that would be greatly appreciated.

ArcGISRuntimeSDKQt_CppSamples/Layers/DisplaySubtypeFeatureLayer/DisplaySubtypeFeatureLayer.cpp
ldanzinger: @jared-esri my concern is calling `m_subtypeFeatureLayer->sublayerWithSubtypeName()` without checking if `m_subtypeFeatureLayer` is nullptr or not first. Wasn't sure if there is any guarantee that this would be not null by the time this was called. If not, I suggest checking if it's null before accessing the sublayerWithSubtypeName

jared-esri: Oh I see now. Yes there is no guarantee it will not be null. I will make the change.

PR #969, opened 2020-02-11
refactor location display samples

ldanzinger:
removing explicit set of compass/positionsource in favor of using the default location data source

@lsmallwood please review


ldanzinger:
@lsmallwood good idea. I double checked and there is nothing related to change

PR #978, opened 2020-02-28
Jcaccamo/config subnet trace

jared-esri:
@JamesMBallard @mfeigl Please review when you have some time. Thank you!


ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    function expressionToString(expression) {
+
+        const traceConditionType = expression.traceConditionType;
+        const networkAttribute = expression.networkAttribute;
+        const otherNetworkAttribute = expression.otherNetworkAttribute;
+        const operatorAsString = comparisonOperatorToString(expression.comparisonOperator)
+
+        switch (traceConditionType) {
```
mfeigl: Please add a default case.

ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
```diff
@@ -0,0 +1,423 @@
...
+            traceConfiguration.traversability.barriers = combineExpressions;
+            expressionBuilder.text = expressionToString(combineExpressions);
+        } else {
+            traceConfiguration.traversability.barriers = expression;
+            expressionBuilder.text = expressionToString(expression);
+        }
+    }
+
+    function convertToDataType(data, dataType) {
+        switch(dataType) {
```
mfeigl: Would a default clause make sense in here?

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+  connect(m_utilityNetwork, &UtilityNetwork::doneLoading, this,&ConfigureSubnetworkTrace::onUtilityNetworkLoaded);
+
+  connect(m_utilityNetwork, &UtilityNetwork::traceCompleted, this, &ConfigureSubnetworkTrace::onTraceCompleted);
+
+  m_utilityNetwork->load();
+}
+
+QString ConfigureSubnetworkTrace::expressionToString(UtilityTraceConditionalExpression* expression) const
+{
+  switch (expression->traceConditionType())
```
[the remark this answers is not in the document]
jared-esri: Good point! I will go back and add default clauses. 👍

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+  m_busy(true),
+  m_dialogVisible(false),
+  m_textFieldVisible(true),
+  m_assetGroupName(QString("Circuit Breaker")),
+  m_assetTypeName(QString("Three Phase")),
+  m_deviceTableName(QString("Electric Distribution Device")),
+  m_domainNetworkName(QString("ElectricDistribution")),
+  m_tierName(QString("Medium Voltage Radial")),
+  m_featureLayerUrl(QUrl("https://sampleserver7.arcgisonline.com/arcgis/rest/services/UtilityNetwork/NapervilleElectric/FeatureServer")),
+  m_gloabId(QUuid("{1CAF7740-0BF4-4113-8DB2-654E18800028}"))
```
JamesMBallard: Optional, but I would recommend initializing these all in the header.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+ConfigureSubnetworkTrace::ConfigureSubnetworkTrace(QObject* parent /* = nullptr */):
+  QObject(parent),
+  m_busy(true),
+  m_dialogVisible(false),
+  m_textFieldVisible(true),
+  m_assetGroupName(QString("Circuit Breaker")),
+  m_assetTypeName(QString("Three Phase")),
+  m_deviceTableName(QString("Electric Distribution Device")),
+  m_domainNetworkName(QString("ElectricDistribution")),
```
JamesMBallard: Initializing QString with `(QString("someText"))` is redundant.

```suggestion
  m_domainNetworkName("ElectricDistribution"),
```

Same suggestion for all of these.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.h
```diff
@@ -0,0 +1,115 @@
...
+
+private slots:
+  void onTraceCompleted();
+  void onUtilityNetworkLoaded(const Esri::ArcGISRuntime::Error& e);
+
+private:
+
+  QString comparisonOperatorToString(const Esri::ArcGISRuntime::UtilityAttributeComparisonOperator& comparisonOperator) const;
+  QString expressionToString(Esri::ArcGISRuntime::UtilityTraceConditionalExpression* expression) const;
+  QVariant convertToDataType(const QVariant& value, const Esri::ArcGISRuntime::UtilityNetworkAttributeDataType& dataType);
```
JamesMBallard: Optional, but I don't think any of these actually need to be member functions. They are all stateless, so could easily be static helpers.

Possibly not `convertToDataType` since it is not const, but unsure about that.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+  const QVariant convertedSelectedValue = convertToDataType(selectedValue, selectedNetworkAttribute->dataType());
+
+  const UtilityAttributeComparisonOperator selectedOperatorEnum = static_cast<UtilityAttributeComparisonOperator>(selectedOperator);
+
+
+  // NOTE: You may also create a UtilityNetworkAttributeComparison with another NetworkAttribute.
+  UtilityTraceConditionalExpression* expression = new UtilityNetworkAttributeComparison(selectedNetworkAttribute, selectedOperatorEnum, convertedSelectedValue, this);
+
+  if (traversability->barriers())
```
JamesMBallard: This could be a problem. Above, you _might_ execute `m_traceConfiguration->setTraversability(new UtilityTraversability(this));`, so this could be `nullptr` in that case.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+  if (!traversability)
+    m_traceConfiguration->setTraversability(new UtilityTraversability(this));
+
+  UtilityNetworkAttribute* selectedNetworkAttribute = m_networkDefinition->networkAttribute(selectedAttribute);
+
+  const QVariant convertedSelectedValue = convertToDataType(selectedValue, selectedNetworkAttribute->dataType());
+
+  const UtilityAttributeComparisonOperator selectedOperatorEnum = static_cast<UtilityAttributeComparisonOperator>(selectedOperator);
+
+
```
JamesMBallard: No need for two empty lines here.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+void ConfigureSubnetworkTrace::changeIncludeContainersState(bool includeContainers)
+{
+  m_traceConfiguration->setIncludeContainers(includeContainers);
+}
+
+void ConfigureSubnetworkTrace::reset()
+{
+  // Reset the barrier condition to the initial value.
+  m_traceConfiguration->traversability()->setBarriers(m_initialExpression);
```
JamesMBallard: There is code above that checks if `m_traceConfiguration->traversability()` is valid. Can we assume here that it is always valid?

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+  emit expressionBuilderChanged();
+}
+
+void ConfigureSubnetworkTrace::trace()
+{
+  m_busy = true;
+  emit busyChanged();
+  if (!m_utilityNetwork || !m_utilityElementStartingLocation)
+  {
+    return;
```
JamesMBallard: Is ` m_busy = true;` still correct if we do any early return here? I'm wondering if that will cause a progress spinner to active with nothing in progress.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+    // trace the network
+    m_utilityNetwork->trace(m_traceParams);
+  }
+}
+
+void ConfigureSubnetworkTrace::onTraceCompleted()
+{
+  if (m_utilityNetwork->traceResult()->isEmpty())
+  {
+    m_dialogText = QString("No results returned");
```
JamesMBallard:
```suggestion
    m_dialogText = "No results returned";
```

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+  }
+  emit attributeListModelChanged();
+
+  // Create a default starting location.
+  const UtilityNetworkSource* networkSource = m_networkDefinition->networkSource(m_deviceTableName);
+  const UtilityAssetGroup* assetGroup = networkSource->assetGroup(m_assetGroupName);
+  UtilityAssetType* assetType = assetGroup->assetType(m_assetTypeName);
+  m_utilityElementStartingLocation = m_utilityNetwork->createElementWithAssetType(assetType, m_gloabId);
+
+  QList<UtilityTerminal*> terminals = m_utilityElementStartingLocation->assetType()->terminalConfiguration()->terminals();
```
JamesMBallard: Is this entire call chain guaranteed to be valid?

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+  // Set the terminal for this location. (For our case, we use the 'Load' terminal.)
+  for (UtilityTerminal* terminal : terminals)
+  {
+    if (terminal->name() == "Load")
+    {
+      m_utilityElementStartingLocation->setTerminal(terminal);
+      break;
+    }
+  }
```
JamesMBallard: Optional, consider [std::find_if](https://en.cppreference.com/w/cpp/algorithm/find) here. If that results in even more code or more complicated code - keep what you have.

ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
```diff
@@ -0,0 +1,423 @@
...
+            traceConfiguration.traversability.scope = Enums.UtilityTraversabilityScopeJunctions;
+
+            expressionBuilder.text = expressionToString(initialExpression);
+        }
+
+        onComponentCompleted: load();
+    }
+
+    function updateInputMethod(currentText, currentIndex) {
+        if(networkDefinition) {
```
JamesMBallard:
```suggestion
        if (networkDefinition) {
```

ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
```diff
@@ -0,0 +1,423 @@
...
+
+                inputTextField.visible = false;
+            } else {
+                inputTextField.visible = true;
+            }
+        }
+    }
+
+    function expressionToString(expression) {
+
```
JamesMBallard: No need to begin a function with an empty line.

ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
```diff
@@ -0,0 +1,423 @@
...
+    id: rootRectangle
+    clip: true
+    width: 800
+    height: 600
+
+    property var attributeStringListModel: []
+    property var initialExpression
+    property var networkDefinition
+    property var traceConfiguration
+    property var utilityElementStartingLocation
```
JamesMBallard: Should all of these be initialized to `null`?

ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
```diff
@@ -0,0 +1,423 @@
...
+            return "DoesNotIncludeAny";
+        default:
+            return null;
+        }
+    }
+
+    function addCondition() {
+        // NOTE: You may also create a UtilityCategoryComparison with UtilityNetworkDefinition.Categories and UtilityCategoryComparisonOperator.
+
+        if (traceConfiguration === null) {
```
JamesMBallard:
```suggestion
        if (!traceConfiguration) {
```
Optional.

Also, is `traceConfiguration` initialized to null above?

ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
```diff
@@ -0,0 +1,423 @@
...
+            traceConfiguration = ArcGISRuntimeEnvironment.createObject("UtilityTraceConfiguration");
+        }
+        if (traceConfiguration.traversability === null) {
```
JamesMBallard:
```suggestion
        if (!traceConfiguration.traversability) {
```

JamesMBallard: Also, since this is a standalone clause, I would put a line between this if-statement and the previous one. When they're grouped like this my mind assumes the second one is an else-if and they are connected.

ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
```diff
@@ -0,0 +1,423 @@
...
+            traceConfiguration.traversability = ArcGISRuntimeEnvironment.createObject("UtilityTraversability");
+        }
+
+        const selectedNetworkAttribute = networkDefinition.networkAttribute(networkAttributeComboBox.currentText);
+        const data = convertToDataType(valueSelectionComboBox.visible ? valueSelectionComboBox.currentIndex : inputTextField.text, selectedNetworkAttribute.dataType)
+
+        if (isNaN(data)) {
```
JamesMBallard: This is a bit confusing to me. This is the result of `convertToDataType`, and I would not think that a "type" could be `NAN`. I would suggest a minor cleanup here to clarify this.

ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
```diff
@@ -0,0 +1,423 @@
...
+
+        // NOTE: You may also create a UtilityNetworkAttributeComparison with another NetworkAttribute.
+        const expression = ArcGISRuntimeEnvironment.createObject("UtilityNetworkAttributeComparison", {
+                                                                     networkAttribute: selectedNetworkAttribute,
+                                                                     comparisonOperator: comparisonOperatorComboBox.currentIndex,
+                                                                     value: data});
+
+        const otherExpression = traceConfiguration.traversability.barriers;
+
+        if (otherExpression) {
```
JamesMBallard: Seems confusing to read this. `otherExpression` = `barriers`, which sounds like a list, and a list cannot be null.

ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
```diff
@@ -0,0 +1,423 @@
...
+
+    function reset() {
+        traceConfiguration.traversability.barriers = initialExpression;
+        expressionBuilder.text = expressionToString(initialExpression);
+    }
+
+    function trace() {
+        busyIndicator.visible = true;
+        if (!utilityNetwork || !utilityElementStartingLocation) {
+            return;
```
JamesMBallard: Same comment as C++. If you return early here, `busyIndicator.visible = true;` is probably wrong. You may want to put that in the else block.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+  emit busyChanged();
+  m_networkDefinition = m_utilityNetwork->definition();
+
+  // Build the choice lists for network attribute comparison.
+  for (UtilityNetworkAttribute* networkAttribute : m_networkDefinition->networkAttributes())
+  {
+    if (!networkAttribute->isSystemDefined())
+      m_attributeListModel.append(networkAttribute->name());
+  }
+  emit attributeListModelChanged();
```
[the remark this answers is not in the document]
jared-esri: > Shouldn't this be right beneath `m_attributeListModel.append(networkAttribute->name());`?

I only emit the signal once I know the list is done being populated rather than every time one is added.

> What if all of the network attributes are system defined?

In the case of this sample we know that we want to use the attributes that are not system defined. If the user inputs their own data we expect them to make the other necessary changes as well.

Specifically for this sample though, system defined attributes could work but don't make a whole lot of sense in the context of the sample. E.g Tier name or Source ID.

ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
jared-esri: I agree that is definitely confusing. I could add a comment explaining that this is checking to make sure that this is not the first expression being added to the barriers condition.

After thinking about this a little, I think it would be better to remove this though. I had added some code here and there in the case the user adds in their own data. Our general stance on this is that the sample should work as is and if the user changes the data, it's expected they could modify the code as well, if needed. This checks for an edge case that isn't expected in this sample.

JamesMBallard: That would be best if you can remove this. I wouldn't expect the sample code to handle any data. It is tuned for the dataset it is written against.

ArcGISRuntimeSDKQt_QMLSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.qml
jared-esri: Good point. In the case no input was entered, it would end up returning 'nan' so this was a quick fix for that. I think my change is a better approach to a situation with invalid input.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
jared-esri: Yes `m_traceConfiguration` and `m_traceConfiguration->traversability()` is expected to always be valid. These can both be removed from `addCondition` for both C++ and QML.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+}
+
+void ConfigureSubnetworkTrace::onTraceCompleted()
+{
+  if (m_utilityNetwork->traceResult()->isEmpty())
+  {
+    m_dialogText = QString("No results returned");
+    emit dialogTextChanged();
+    m_dialogVisible = true;
+    emit dialogVisibleChanged();
```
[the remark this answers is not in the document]
jared-esri: Thank you for forcing me to look at this a little more closely. I followed a previous implementation of this but I it's been bothering me for a while. I think just having a signal , `void showDialog()` to open the dialog makes it much more clear and to the point.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,368 @@
...
+  {
+    m_dialogText = QString("%1 - %2").arg(e.message(), e.additionalMessage());
+    m_dialogVisible = true;
+    emit dialogTextChanged();
+    emit dialogVisibleChanged();
+    return;
+  }
+
+  m_busy = false;
+  emit busyChanged();
```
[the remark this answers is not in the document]
jared-esri: I think I have these situations covered where `m_busy` just flips from true to false and it shouldn't be set to itself.

ArcGISRuntimeSDKQt_CppSamples/Analysis/ConfigureSubnetworkTrace/ConfigureSubnetworkTrace.cpp
```diff
@@ -0,0 +1,353 @@
...
+#include "UtilityTraceResultListModel.h"
+#include "UtilityTraversability.h"
+
+#include <QQmlEngine>
+#include <algorithm>
+
+using namespace Esri::ArcGISRuntime;
+QString expressionToString(UtilityTraceConditionalExpression* expression);
+QString comparisonOperatorToString(const UtilityAttributeComparisonOperator& comparisonOperator);
+QVariant convertToDataType(const QVariant& value, const Esri::ArcGISRuntime::UtilityNetworkAttributeDataType& dataType);
```
JamesMBallard: My mistake, I meant these should remain class functions but become static class functions. They all seemed like helpers that don't need access to any class members.

PR #1017, opened 2020-03-18
Added clipping distance to tabletop AR sample.

jared-esri:
@ldanzinger @GuillaumeBelz Ready for review. Thank you.

I saw that this was almost done/is done so I figured I would just make the small change and have it ready to go.


GuillaumeBelz:
> I saw that this was almost done/is done

Yes, this is merged.

PR #1022, opened 2020-03-23
Update ConfigureSubnetworkTrace.cpp

ldanzinger:
@jared-esri please review. Seems like we should return here if no result.


ldanzinger:
Closing this. I will add this fix to another PR.

PR #1030, opened 2020-03-25
update screenshot.

jared-esri:
@ldanzinger ready for review. Thanks!

Screenshot needed to be updated since the addition of clipping distance changes how the scene is displayed.


ldanzinger:
@jared-esri screenshots should be 500x500. Ideally they wouldn't have someone's leg in it either... Can it be clipped or will it be too small?

jared-esri:
> @jared-esri screenshots should be 500x500. Can it be clipped or will it be too small?

If i had a tablet with good resolution I could manage 500x500 but with a phone the width is around 300 which makes it difficult and I think if it was clipped it would be too small. 

> Ideally they wouldn't have someone's leg in it either...

Agreed. I don't have a suitable location to take the screenshot currently and I just made use of the .Net screenshot since there is no visual difference when running the sample.

I might be able to come up with a location for the screenshot that won't include feet. The dimensions might be a bit more tricky since I would have to enlarge the image for it to fit the proportions which will likely cause the image to degrade. Not exactly sure how much but it's worth a shot. I am by no means an expert on this so take my words with a grain of salt.

jared-esri:
Taking a screenshot on the actual device and not a screenshot from the computer solved the dimensions issue so it was easier to get the 500x500 and no more feet!

PR #1046, opened 2020-04-08
Update NavigateRoute.qrc

ldanzinger:
@JamesMBallard please review. While testing I noticed these files were missing from QRC which resulted in the h/cpp files not showing up in the viewer


ldanzinger:
cc @sbiswas96

PR #1075, opened 2020-05-21
Jared/update imageoverlay json

jared-esri:
@sbiswas96 @JamesMBallard Ready for review. Thanks!

I think I got all the pieces needed for this to work properly and removed what was no longer needed. I tested both sampleviewers and had no issues besides C++ which I had to reorganize the files in snippets for it to load, not sure why that is but it was a trend I noticed with others.


sbiswas96:
@jared-esri the issue you noticed about needing to rearrange the files in the "snippets" section is due to needing the qml file for the sample to be listed first, because the viewer uses the first file in the "snippets" section to initialize the viewer. The updated templates should automatically put the qml file first.

PR #1078, opened 2020-05-28
Sbiswas/kml contents

sbiswas96:
Ready for review when you get a chance, thanks.

Sample design [here](https://devtopia.esri.com/runtime/common-samples/tree/master/designs/ListKmlContents). This sample loads a KML layer and outputs the hierarchy of nodes in a list of clickable buttons. When you click a button, the scene zooms to that node's extent and the list of buttons updates to display the node's children, if applicable.


ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -0,0 +1,320 @@
...
+  connect(m_kmlDataset, &KmlDataset::doneLoading, this, [this](Error loadError)
+  {
+    if (!loadError.isEmpty())
+    {
+      qDebug() << loadError.message() << loadError.additionalMessage();
+      return;
+    }
+
+    // recursively build tree to display KML contents
+    for (int i = 0; i < m_kmlDataset->rootNodes().size(); ++i)
```
lsmallwood, line 83: can we do a range loop here?

```suggestion
    for (const KmlNode* node : qAsConst(m_kmlDataset->rootNodes()))
```

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -0,0 +1,320 @@
...
+    if (!m_kmlNodesList.isEmpty() && m_kmlNodesList[0]->parentNode() == nullptr)
+    {
+      displayChildren(m_kmlNodesList[0]);
+    }
+  });
+}
+
+QString ListKmlContents::getKmlNodeType(KmlNode *node)
+{
+  QString type = "";
```
lsmallwood, line 101: should check `node` for `nullptr`

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -0,0 +1,320 @@
...
+  case KmlNodeType::KmlTour:
+    type = "KmlTour";
+    break;
+  }
+  return " - " + type;
+}
+
+void ListKmlContents::displayChildren(KmlNode *parentNode)
+{
+  if (KmlContainer* container = dynamic_cast<KmlContainer*>(parentNode))
```
lsmallwood, line 133: check ptr for nullptr

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -0,0 +1,320 @@
...
+      m_currentNode = container->childNodesListModel()->at(0);
+    }
+    emit levelNodeNamesChanged();
+  }
+}
+
+// recursively build string to indicate node's ancestors
+void ListKmlContents::buildPathLabel(KmlNode* node)
+{
+  if (node->parentNode() != nullptr)
```
lsmallwood, line 163: check ptr for nullptr

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -0,0 +1,320 @@
...
+    buildPathLabel(node->parentNode());
+    m_labelText.append(">");
+  }
+  m_labelText.append(node->name());
+}
+
+void ListKmlContents::displayPreviousLevel()
+{
+  KmlNode* parentNode = m_currentNode->parentNode();
+  KmlNode* grandparentNode = parentNode->parentNode();
```
lsmallwood, line 174: check parentNode

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -0,0 +1,320 @@
...
+  }
+}
+
+// display selected node on sceneview and show its children
+void ListKmlContents::nodeSelected(QString nodeName)
+{
+  // find node in the list
+  for (KmlNode* node : m_kmlNodesList)
+  {
+    if (nodeName.contains(" - "))
```
lsmallwood, line 210: this seems a little bit fragile if people can name their own node...

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+  if (m_currentNode->name() == "")
+  {
+    m_isTopLevel = true;
+    emit isTopLevelChanged();
+  }
+}
+
+// display selected node on sceneview and show its children
+void ListKmlContents::nodeSelected(QString nodeName)
```
lsmallwood, line 205: looks like `nodeName` should be passed by const&

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.h
```diff
@@ -0,0 +1,89 @@
...
+{
+class KmlDataset;
+class KmlNode;
+class KmlNodeListModel;
+class Scene;
+class SceneQuickView;
+}
+}
+
+//class QAbstractListModel;
```
lsmallwood, line 32: remove this?

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.qml
```diff
@@ -0,0 +1,99 @@
...
+                            width: textRectangle.width
+                            wrapMode: Text.Wrap
+                        }
+                    }
+                }
+
+                RowLayout {
+                    ListView {
+                        id: myListView
+//                        anchors.horizontalCenter: parent.horizontalCenter
```
lsmallwood, line 73: not required?

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/README.md
```diff
@@ -0,0 +1,50 @@
...
+## How to use the sample
+
+The contents of the KML file are shown in a list of buttons. Select a node to zoom to that node and see its child nodes, if it has any. Not all nodes can be zoomed to (e.g. screen overlays).
+
+## How it works
+
+1. Add the KML file to the scene as a layer.
+2. Explore the root nodes of the `KmlDataset` recursively to create a view model.
+  * Each node is enabled for display at this step. KML files may include nodes that are turned off by default.
+3. When a node is selected, use the node's `Extent` to determine a viewpoint and set the `SceneView` object's viewpoint do it.
```
lsmallwood, line 20:
```suggestion
3. When a node is selected, use the node's `Extent` to determine a viewpoint and set the `SceneView` object's viewpoint to it.
```

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/README.md
```diff
@@ -0,0 +1,50 @@
...
+* KmlPlacemark
+* KmlScreenOverlay
+
+## Offline data
+
+Read more about how to set up the sample's offline data [here](http://links.esri.com/ArcGISRuntimeQtSamples).
+
+Link | Local Location
+---------|-------|
+|[ESRI test data](https://www.arcgis.com/home/item.html?id=da301cb122874d5497f8a8f6c81eb36e)| `<userhome>`/ArcGIS/Runtime/Data/kml/esri_test_data.kmz |
```
lsmallwood, line 41: I feel like when we use a local file in samples, the code that uses it (e.g. the ctor in your code above) should check for the file's existence and direct the user to these resources if it's missing. (Not required for this change)

ArcGISRuntimeSDKQt_QMLSamples/Layers/ListKmlContents/ListKmlContents.qml
```diff
@@ -0,0 +1,278 @@
...
+    function displayChildren(parentNode) {
+        // if node has children, then display children
+        if (parentNode.childNodesListModel !== null && parentNode.childNodesListModel !== undefined) {
+            let childNodes = parentNode.childNodesListModel
+            let lastLevel = true;
+
+            // clear previous node names
+            nodesOnLevel = [];
+            for (let i = 0; i < childNodes.count; i++) {
+                nodesOnLevel.push(childNodes.get(i).name + getKmlNodeType(childNodes.get(i)));
```
lsmallwood, line 70: call `childNodes.get(i)` once and cache it?

ArcGISRuntimeSDKQt_QMLSamples/Layers/ListKmlContents/ListKmlContents.qml
```diff
@@ -0,0 +1,278 @@
...
+                            buildTree(rootNodes[i]);
+                        }
+
+                        // if at top node, then display children
+                        if (kmlNodesList.length !== 0 && kmlNodesList[0].parentNode === null) {
+                            displayChildren(kmlNodesList[0]);
+                        }
+                    }
+
+
```
lsmallwood, line 273: extra empty line

ArcGISRuntimeSDKQt_QMLSamples/Layers/ListKmlContents/README.md
```diff
@@ -0,0 +1,48 @@
...
+## How to use the sample
+
+The contents of the KML file are shown in a list of buttons. Select a node to zoom to that node and see its child nodes, if it has any. Not all nodes can be zoomed to (e.g. screen overlays).
+
+## How it works
+
+1. Add the KML file to the scene as a layer.
+2. Explore the root nodes of the `KmlDataset` recursively to create a view model.
+  * Each node is enabled for display at this step. KML files may include nodes that are turned off by default.
+3. When a node is selected, use the node's `Extent` to determine a viewpoint and set the `SceneView` object's viewpoint do it.
```
lsmallwood, line 20:
```suggestion
3. When a node is selected, use the node's `Extent` to determine a viewpoint and set the `SceneView` object's viewpoint to it.
```

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
sbiswas96: `nodeName` is manipulated with
`nodeName = nodeName.left(ind);` which is why I passed by value

lsmallwood: but this is doing a copy constructor right? you might as well do:

`QString newNodeName = nodeName.left(ind);`

sbiswas96: Ohh right! I see

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.qml
```diff
@@ -33,7 +33,10 @@ Item {
             height: childrenRect.height
             color: "lightgrey"
 
-            ColumnLayout {
+            GridLayout {
+                columns: 1
```
anmacdonald, line 37: Columns should be 2.

Remove the `RowLayout`s. `myListView` should just span 2 columns using the injected property `Layout.columnSpan`.


Remove all instances of manually set width and height. Use `Layout.fillWidth` on the textRectangle.

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.qml
```diff
@@ -33,7 +33,10 @@ Item {
...
+                rows: 2
```
anmacdonald, line 38: Don't need to specify rows.

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -313,22 +288,26 @@ void ListKmlContents::setSceneView(SceneQuickView* sceneView)
...
-QStringList ListKmlContents::levelNodeNames() const
+QStringList ListKmlContents::levelNodeNames()
 {
   return m_levelNodeNames;
 }
 
-QString ListKmlContents::labelText() const
+QString ListKmlContents::labelText()
 {
+  m_labelText.clear();
```
anmacdonald, line 298: You don't need to keep `m_labelText` to do this recurisve function.

Try:


```cpp
// recursively build string to indicate node's ancestors
QStringList ListKmlContents::buildPathLabel(KmlNode* node) const
{
  if (node)
    return buildPathLabel(node->parentNode()) << node->name();
  else
    return QStringList { };
}
```

With:

```cpp
QString ListKmlContents::labelText() const 
{
  return buildPathLabel(m_currentNode).join(" > ");
}
```

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -130,10 +129,10 @@ QString ListKmlContents::getKmlNodeType(KmlNode *node)
   default:
     return "";
   }
-  return " - " + type;
+  return type;
 }
 
-void ListKmlContents::displayChildren(KmlNode *parentNode)
```
anmacdonald: If we are just grabbing the children of the current node, I feel this does not need to be cached.

Again, this function is doing some surprising behaviour, we're wanting to list all the children for the given node, but this function also has logic to change `m_currentNode`. This does not feel like the remit of `displayChildren`?

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -0,0 +1,313 @@
...
+  {
+    buildPathLabel(node->parentNode());
+    m_labelText.append(">");
+  }
+  m_labelText.append(node->name());
+}
+
+void ListKmlContents::displayPreviousLevel()
+{
+  KmlNode* parentNode = m_currentNode->parentNode();
```
anmacdonald, line 184: Intuitively, it makes sense to me that clicking `displayPreviousLevel` should only jump back by one node. Why do we try and skip back by 2?

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -0,0 +1,313 @@
...
+{
+  QString extractedNodeName = nodeName;
+  if (nodeName.contains(" - "))
+  {
+    int ind = nodeName.lastIndexOf(" - ");
+    extractedNodeName = nodeName.left(ind);
+  }
+
+  // find node in the list
+  for (KmlNode* node : m_kmlNodesList)
```
anmacdonald, line 218: I don't think that `m_kmlNodesList` needs to exist. We just need to loop through the immediate children of `m_curretNode`.

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+// display selected node on sceneview and show its children
+void ListKmlContents::processSelectedNode(const QString& nodeName)
+{
+  QString extractedNodeName = nodeName;
+  if (nodeName.contains(" - "))
+  {
+    int ind = nodeName.lastIndexOf(" - ");
+    extractedNodeName = nodeName.left(ind);
+  }
```
anmacdonald, line 215: It took me a while to realize what was going on here. 

You don't necessarilly need to do any string manipulation to match the names.

You could just do:

```cpp
 if (nodeName.startsWith(node->name()))
 {
...
```

sbiswas96: I don't think this will work in cases where 2 names begin in the same way. For example, we have "Building O" and "Building OA" as 2 different nodes

anmacdonald: Ahh good call

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
sbiswas96: It is possible to click on different nodes on the same level, i.e., siblings.
image

For example in the image above, I could click on "Simple placemark" and `m_currentNode` would correspond to the "Simple placemark" node. I should then be able to click on "Floating placemark" but if I just searched through the children of `m_currentNode` I wouldn't find the "Floating placemark" node.

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
sbiswas96: You go back 2 levels to the grandparent and then display its children, so `m_currentNode`'s parent and its siblings

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
anmacdonald: So the list is a combination of current children and siblings?

sbiswas96: The list is built once using `buildTree` and contains all the nodes in the tree

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -189,24 +154,25 @@ void ListKmlContents::displayPreviousLevel()
 
   if (grandparentNode != nullptr)
   {
-    displayChildren(grandparentNode);
     m_currentNode = grandparentNode;
     emit currentNodeChanged();
     emit labelTextChanged();
```
anmacdonald, line 159: Does the `labelTextChanged` signal need to exist or can it be replaced by `currentNodeChanged` now?

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.h
```diff
@@ -74,7 +73,7 @@ class ListKmlContents : public QObject
   QStringList m_levelNodeNames = {};
   QList<Esri::ArcGISRuntime::KmlNode*> m_kmlNodesList = {};
   Esri::ArcGISRuntime::KmlNode* m_currentNode = nullptr;
-  QString m_labelText = "";
+  bool m_lastLevel;
```
anmacdonald, line 76: I think functions like `displayPreviousLevel()` shout be setting this but don't. I'd suggest making a `bool hasDescendents(KmlNode* node)` function instead to remove the need for caching a value.

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
anmacdonald: The same comment applies for `labelTextChanged`.

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -231,8 +197,16 @@ void ListKmlContents::processSelectedNode(const QString& nodeName)
...
-      // show the children of the node
-      displayChildren(node);
+      // reset m_lastLevel before levelNodeNames() is called
+      m_lastLevel = false;
+      emit levelNodeNamesChanged();
+
+      // if displaying end-nodes, change m_currentNode to first end-node for correct behavior of back button
+      if (m_lastLevel)
+      {
+        m_currentNode = dynamic_cast<KmlContainer*>(m_currentNode)->childNodesListModel()->at(0);
```
anmacdonald, line 207: If we can guarantee this is a container, we can use `static_cast` which is faster at runtime. Otherwise we should check if the cast returns a `nullptr` or not!

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
sbiswas96: both those signals are still needed because they are not always called whenever `currentNodeChanged` is emitted

anmacdonald: I'm unable to find a case where `emit labelTextChanged();` is not called immedialtey after `emit currentNodeChanged();`.

sbiswas96: Sometimes `emit currentNodeChanged()` is called but `emit labelTextChanged()` is not called

PR #1093, opened 2020-06-08
Sbiswas/feature collection portal

sbiswas96:
Ready for review, thanks! I am simplifying the code so the portal item is hardcoded and automatically loaded upon the sample being launched instead of asking the user for input.


ArcGISRuntimeSDKQt_CppSamples/Layers/FeatureCollectionLayerFromPortal/FeatureCollectionLayerFromPortal.cpp
```diff
@@ -28,61 +28,36 @@
...
+  m_map(new Map(Basemap::oceans(this), this))
 {
-  if (m_portalItem != nullptr && m_portalItem->loadStatus() == LoadStatus::Loading)
-  {
-    m_messageText = "Portal item loading in progress";
-    emit messageTextChanged();
-    return;
-  }
+  Portal* portal = new Portal(this);
+  m_portalItem = new PortalItem(portal, featureCollectionItemId, this);
```
lsmallwood, line 41: you could move this into the initializer list (e.g. create the portal in-line)

PR #1103, opened 2020-06-17
Sbiswas/edit list kml contents2

sbiswas96:
Thanks for reviewing! 

With the previous implementation, the viewpoint is not set correctly for the nodes because elevation was not taken into account. Additionally, due to [ft-pin issue #168](https://devtopia.esri.com/runtime/ft-ping/issues/168) the `Camera` is set manually with `setViewpointCameraAndWait` to follow the workaround presented in the issue.


ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -210,6 +215,95 @@ void ListKmlContents::processSelectedNode(const QString& nodeName)
...
+
+  if (!kmlViewpoint.isEmpty())
+  {
+    // altitude adjustment is needed for all but Absolute altitude mode
+    m_needsAltitudeFixed = (kmlViewpoint.altitudeMode() != KmlAltitudeMode::Absolute);
+
+    switch (kmlViewpoint.type())
+    {
+    case KmlViewpointType::LookAt:
+      m_viewpoint = Viewpoint(kmlViewpoint.location(), Camera(kmlViewpoint.location(), kmlViewpoint.range(),
```
lsmallwood, line 230: I guess you could shorten these 2 case statements if you had create the camera before hand... looks the same in both scenarios?

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -210,6 +215,95 @@ void ListKmlContents::processSelectedNode(const QString& nodeName)
...
+  // if viewpoint was empty, then use node's extent
+  const Envelope nodeExtent = node->extent();
+  if (nodeExtent.isValid() && !nodeExtent.isEmpty())
+  {
+    // when no altitude is specified, assume elevation needs to be adjusted
+    m_needsAltitudeFixed = true;
+
+    if (nodeExtent.width() == 0 && nodeExtent.height() == 0)
+    {
+      // default values based on Google Earth
```
lsmallwood, line 252: not sure if we should say this?

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -210,6 +215,95 @@ void ListKmlContents::processSelectedNode(const QString& nodeName)
...
+  }
+
+  // if altitude mode is Absolute, viewpoint doesn't need adjustment
+  if (altMode == KmlAltitudeMode::Absolute)
+  {
+    m_sceneView->setViewpoint(m_viewpoint);
+    return;
+  }
+
+  const Envelope lookAtExtent = static_cast<Envelope>(m_viewpoint.targetGeometry());
```
lsmallwood, line 294: you should probably use the `geometry_cast` helper here instead. And I think that we can't use the `targetGeometry` for both. Maybe the envelope should be `targetGeometry().extent()`?

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -257,6 +351,98 @@ void ListKmlContents::setSceneView(SceneQuickView* sceneView)
...
+    if (!kmlViewpoint.isEmpty())
+    {
+      altMode = kmlViewpoint.altitudeMode();
+    }
+    // if altitude mode is Absolute, viewpoint doesn't need adjustment
+    if (altMode == KmlAltitudeMode::Absolute)
+      return;
+
+    Envelope lookAtExtent = static_cast<Envelope>(m_viewpoint.targetGeometry());
+    Point lookAtPoint = static_cast<Point>(m_viewpoint.targetGeometry());
```
lsmallwood, line 369: same comments about casting

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -257,6 +351,98 @@ void ListKmlContents::setSceneView(SceneQuickView* sceneView)
...
+                          lookAtExtent.spatialReference());
+      }
+
+      // if node has viewpoint specified, return adjusted geometry with adjusted camera
+      if (!kmlViewpoint.isEmpty())
+      {
+        m_sceneView->setViewpointCameraAndWait(m_viewpoint.camera().elevate(elevation));
+        m_viewpoint = Viewpoint(target);
+        m_sceneView->setViewpoint(m_viewpoint);
+        return ;
```
lsmallwood, line 406:
```suggestion
        return;
```

ArcGISRuntimeSDKQt_QMLSamples/Layers/ListKmlContents/ListKmlContents.qml
```diff
@@ -231,6 +330,111 @@ Rectangle {
...
+                    if (locationToElevationStatus !== Enums.TaskStatusCompleted)
+                        return;
+
+                    // assume altitude mode is clamp-to-ground if not specified
+                    let altMode = Enums.KmlAltitudeModeClampToGround;
+                    let kmlViewpoint = currentNode.viewpoint;
+
+                    if (kmlViewpoint !== undefined && kmlViewpoint !== null) {
+                        altMode = kmlViewpoint.altitudeMode;
+                    }
```
lsmallwood, line 344: there's a bit of a mixture of styles with regards brackets around single line if statements. i'm not sure what our code style for samples is - but good to be consistent within the one file.

ArcGISRuntimeSDKQt_QMLSamples/Layers/ListKmlContents/ListKmlContents.qml
```diff
@@ -231,6 +330,111 @@ Rectangle {
...
+
+                    // if altitude mode is Absolute, viewpoint doesn't need adjustment
+                    if (altMode === Enums.KmlAltitudeModeAbsolute)
+                        return;
+
+                    if (currentViewpoint.viewpointType === Enums.ViewpointTypeBoundingGeometry) {
+                        let lookAtExtent = currentViewpoint.extent;
+                        let target;
```
lsmallwood, line 352: can this be initialized to something? maybe `null`?

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
sbiswas96: The two `Camera` objects are different, as the first has a distance parameter specified and the second doesn't

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
sbiswas96: this comment was specified in the .NET implementation to indicate where the default values come from. Maybe I can change instead so it just says what the default values are? E.g., `\\ use default values: distance = 1000, pitch = 45` @lsmallwood

lsmallwood: sounds good to me

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
sbiswas96: I think it should be as is because otherwise the cast to `Envelope` seems to always be valid, even in cases where it should be cast to a `Point`

PR #1118, opened 2020-07-30
removing component and changing visibility instead.

jared-2016:
@ldanzinger Ready for review. Thanks.

Fixes a problem where swapping out the delegate on a listview is invalidating the global property. Changing workflow to something a bit more readable and resolves the problem.


ArcGISRuntimeSDKQt_QMLSamples/Routing/RouteAroundBarriers/RouteAroundBarriers.qml
```diff
@@ -316,10 +316,10 @@ Rectangle {
                             text: "Hide directions"
                             onClicked: {
                                 if (text === "Hide directions") {
-                                    directionsView.delegate = blankDelegate;
+                                    directionsView.visible = false;
```
ldanzinger, line 319: Optional - you could just bind the ListView's visibility to this buttons text string

jared-2016:
One last look over the most recent changes as it should be good.

PR #1130, opened 2020-10-20
Jared/raster mosaic

jared-2016:
@ldanzinger @GuillaumeBelz Ready for code review. Thanks!

I will hold off on merging this until the final data being used is published and tested. I will also need to update the screenshot as well once that happens but might as well get the rest out of the way while I wait.


ArcGISRuntimeSDKQt_QMLSamples/Layers/ApplyMosaicRuleToRasters/ApplyMosaicRuleToRasters.qml
```diff
@@ -0,0 +1,114 @@
...
+        Rectangle {
+            width: childrenRect.width
+            height: childrenRect.height
+            anchors {
+                bottom: mapView.attributionTop
+                horizontalCenter: mapView.horizontalCenter
+            }
+            color: "lightgrey"
+            opacity: 0.8
+            visible: rasterLayer.loadStatus === Enums.LoadStatusLoaded ? true : false
```
ldanzinger, line 65:
```suggestion
            visible: rasterLayer.loadStatus === Enums.LoadStatusLoaded
```

I think the `? true : false` is implied automatically

ArcGISRuntimeSDKQt_QMLSamples/Layers/ApplyMosaicRuleToRasters/ApplyMosaicRuleToRasters.qml
```diff
@@ -0,0 +1,114 @@
...
+        } else if (ruleString === "LockRaster") {
+            mosaicRule.mosaicMethod = Enums.MosaicMethodLockRaster;
+            mosaicRule.lockRasterIds = ["1","7","12"];
+        }
+        imageServiceRaster.mosaicRule = mosaicRule;
+    }
+
+    // Helper function to reset the mosaic rule
+    function resetMosaicRule() {
+        mosaicRule = null;
```
ldanzinger, line 111: this seems unnecessary given the next line of code

ldanzinger (approved):
LGTM - a few cleanup suggestions

ArcGISRuntimeSDKQt_CppSamples/Layers/ApplyMosaicRuleToRasters/ApplyMosaicRuleToRasters.cpp
```diff
@@ -0,0 +1,125 @@
...
+    m_mosaicRule->setLockRasterIds(QList<qint64>{1,7,12});
+  }
+  m_imageServiceRaster->setMosaicRule(m_mosaicRule);
+}
+
+// Helper function to reset the mosaic rule
+void ApplyMosaicRuleToRasters::resetMosaicRule()
+{
+  delete m_mosaicRule;
+  m_mosaicRule = nullptr;
```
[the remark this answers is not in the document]
jared-2016: How about creating the MosaicRule by value when setting the parameters so once it goes out of scope the system cleans it up? Then it would remove the need for that helper function as well as remove a member variable. I pushed the jist of the change so you can see it.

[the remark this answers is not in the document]
JamesMBallard: >How about creating the MosaicRule by value...

No. This can lead to assumptions about how the API internals work.

```c++
m_imageServiceRaster->setMosaicRule(&myMosaicRule);
```

Since that value object will get destructed on the next line, you can make the case it is similar to doing this, and there is no expectation this would work:
```c++
myMosaicRule = new MosaicRule(this);
...
m_imageServiceRaster->setMosaicRule(myMosaicRule);
delete myMosaicRule;
// is this the same as m_imageServiceRaster->setMosaicRule(nullptr); ?
```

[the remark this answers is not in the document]
jared-2016: That's a good point. I reverted that last commit to go back to what it was originally. Thank you for the input!

ArcGISRuntimeSDKQt_QMLSamples/Layers/ApplyMosaicRuleToRasters/ApplyMosaicRuleToRasters.qml
```diff
@@ -0,0 +1,113 @@
...
+                    id: imageServiceRaster
+                    url: "https://sampleserver7.arcgisonline.com/arcgis/rest/services/amberg_germany/ImageServer"
+                }
+                onLoadStatusChanged: {
+                    // Once loaded set the viewpoint to the center of the raster layer
+                    if (loadStatus === Enums.LoadStatusLoaded) {
+                        mapView.setViewpointCenterAndScale(rasterLayer.fullExtent.center, 25000);
+
+                        // Create a mosaic rule and set the mosaic rule property of the image service raster
+                        mosaicRule = ArcGISRuntimeEnvironment.createObject("MosaicRule");
```
GuillaumeBelz, line 49: is it possible to create mosaic rule with declarative code, directly in ImageServiceRaster? In C++ code, the mosaic rule is set before the ImageServiceRaster is loaded.

jared-2016: It's not a default property of `ImageServiceRaster` so I don't think so.

I don't think the order in which that first mosaic rule is set actually matters in relation to it being loaded. It's only initialized to a default `MosaicRule` because within the sample viewer it can pull some cached value for that which will alter which mosiac rule the sample starts with depending on where you left off.

GuillaumeBelz: Sorry for delay to answer.
I'm not sure to understand why it's better to create mosaic rule like that, and not using declarative code. Something like:
```qml
ImageServiceRaster {
    id: imageServiceRaster
    url: "https://sampleserver7.arcgisonline.com/arcgis/rest/services/amberg_germany/ImageServer"
    mosaicRule: MosaicRule {
    }
}
```
                }

jared-2016: @GuillaumeBelz Maybe the misunderstanding is on my part. Normally I don't attempt declarative code when it's not a default property of the QML class. When trying to do it declaratively one of two things happens, 

## First
```
ImageServiceRaster {
    id: imageServiceRaster
    url: "https://sampleserver7.arcgisonline.com/arcgis/rest/services/amberg_germany/ImageServer"
    mosaicRule: MosaicRule {
    }
}
```

this will result in an error `"ImageServiceRaster.mosaicRule" is not available in Esri.ArcGISRuntime 100.10` Which maybe it shouldn't.

## Second
```
ImageServiceRaster {
    id: imageServiceRaster
    url: "https://sampleserver7.arcgisonline.com/arcgis/rest/services/amberg_germany/ImageServer"
    MosaicRule {
    }
}
```

this had no tangible effect that I can see and if used as a replacement for line 49, it will not initialize the `MosaicRule` of the `ImageServiceRaster` which in the sample viewer will have some carryover effect when switching samples.

carry-over-example

GuillaumeBelz: > Normally I don't attempt declarative code when it's not a default property of the QML class. When trying to do it declaratively one of two things happens,

To summary:

If a property is writable, it's possible to assign a value to this property in QML
```
MyComponent {
    myProperty: MyProperty {}
}
```

One property can be "default". In this case, this property is used when the name of the property is not specified in QML. The type of the object must be compatible with the type of the property.
```
// Q_CLASSINFO("DefaultProperty", "myProperty") in C++ class definition

MyComponent {
    MyProperty {} // same as "myProperty: MyProperty {}"
}
```

A special case of default property is `objects`, which take a list of QmlObject. In this case, all the children which are not assigned to a specific property are added to the `objects` property. The setter of `objects` checks the type of the objects and assigns them to the property correspondent to the type of the object.

```
// Q_CLASSINFO("DefaultProperty", "object") in C++ class definition

MyComponent {
    MyProperty {} // assigned to myProperty
    Rectangle {} // assigned to the property which accept Rectangle
}
```

If the child object is not assigned to a property and the setter can't find the correct property, nothing happens.

> # First
> this will result in an error "ImageServiceRaster.mosaicRule" is not available in Esri.ArcGISRuntime 100.10 Which maybe it shouldn't.

`mosaicRule` is writable in `QmlImageServiceRaster` (since 100.9) then the code should be valid. I don't know what is the issue, but this is a bug.

> #Second
> this had no tangible effect that I can see and if used as a replacement for line 49, it will not initialize the MosaicRule of the ImageServiceRaster which in the sample viewer will have some carryover effect when switching samples.

`ImageServiceRaster` have a `objects` default property and `mosaicRule` is writable, then this code should work too. I checked the code of `ImageServiceRaster`, this feature is not supported yet. We should add that.

We can create 2 issues, to fix the bug with `mosaicRule` property and to add the support of `MosaicRule` in `objects` property

 @JamesMBallard @ldanzinger

JamesMBallard: >We can create 2 issues, to fix the bug with mosaicRule property and to add the support of MosaicRule in objects property

@jared-2016 @GuillaumeBelz please work together to get those logged.

One is a trivial fix and the other (add default property support) will take a bit more effort.

jared-2016: Thank you for the added explanation! I understand the problem now. I can log both of the issues.

PR #1131, opened 2020-10-29
Jared/update toolkit

jared-2016:
@JamesMBallard @anmacdonald @ldanzinger Ready for review.

This PR currently makes the minimum changes necessary to update the samples which currently make use of any toolkit items. This should be sufficient as to keep other dependent work moving forward. I tested several individual samples that had different changes, specifically ones that used the toolkit components `AuthenticationView` and `PopupStackView` which had more notable changes. All those tested against the beta branch worked as expected.

I assume that we would like to remove all references to the departing version of the toolkit but didn't want to include that into this PR just incase.


anmacdonald (approved):
Looks good!

Just out of interest, what are the references you are referring to removing later, `ARCGIS_TOOLKIT_IMPORT_PATH_2`?

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/AddItemsToPortal/AddItemsToPortal.qml
```diff
@@ -17,14 +17,14 @@
 import QtQuick 2.6
 import QtQuick.Controls 2.2
 import Esri.Samples 1.0
-import Esri.ArcGISRuntime.Toolkit.Dialogs 100.10
+import Esri.ArcGISRuntime.Toolkit 100.9
```
ldanzinger, line 20: my mistake for not updating the issue, but we now have a beta_update10 branch which is based on 100.10 - https://github.com/Esri/arcgis-runtime-toolkit-qt/tree/beta_update10

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/TokenAuthentication/TokenAuthentication.qml
```diff
@@ -58,7 +58,6 @@ TokenAuthenticationSample {
     // Declare an AuthenticationView
     AuthenticationView {
         anchors.fill: parent
-        authenticationManager: authSample.authenticationManager // set the authenticationManager property (this needs to be registered)
```
ldanzinger, line 61: this is such a big win. Nice job @anmacdonald

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ShowPopup/ShowPopup.qml
```diff
@@ -47,7 +48,7 @@ Item {
         id: model
         mapView: view
 
-        onPopupManagersChanged: popupStackView.show();
```
ldanzinger, line 50: For the sake of portability, I think we should consider adding a show method if it is not present in the current beta_update10 branch. It seems pretty harmless and gets rid of actual code changes

ArcGISRuntimeSDKQt_CppSamples/EditData/UpdateAttributesFeatureService/main.cpp
```diff
@@ -59,6 +61,8 @@ int main(int argc, char *argv[])
   // Add the Toolkit path
   view.engine()->addImportPath(arcGISToolkitImportPath);
 
+  Esri::ArcGISRuntime::Toolkit::registerComponents();
```
ldanzinger, line 64: This is also really cool @anmacdonald. So much better than before

ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMapLocalBasemap/GenerateOfflineMapLocalBasemap.qml
```diff
@@ -152,7 +152,6 @@ GenerateOfflineMapLocalBasemapSample {
     /* Uncomment this section when running as standalone application
     AuthenticationView {
         anchors.fill: parent
-        authenticationManager: offlineMapSample.authenticationManager // set the authenticationManager property (this needs to be registered)
```
ldanzinger, line 155: @anmacdonald @JamesMBallard - Just a thought - removing this property is great and simplifies the code. It also seems to be the most common recurring code change in the PR - everything else is related to imports and registration. Should we consider adding a dummy no-op property on AuthenticationView called authenticationManager just for portability? Everything will work whether you set it or not, but it won't require code change.

My initial thought is no, because the QML Engine will just fail to load the QML file if you try to set a non existent property, so it should be pretty obvious and easy to change this.

ldanzinger:
@jared-2016 won't we also need to update each sample's .pro file to include the new PRI?

jared-2016:
> Looks good!
> 
> Just out of interest, what are the references you are referring to removing later, `ARCGIS_TOOLKIT_IMPORT_PATH_2`?

@anmacdonald Yes that's one and the other is `ARCGIS_TOOLKIT_IMPORT_PATH` which is used for some linux platform replacement.

See here for an exmaple, https://github.com/Esri/arcgis-runtime-samples-qt/blob/8252c8205325f0b5ce9c1e3145d3ba8c6da744d3/ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/PortalUserInfo/main.cpp#L52-L61.

jared-2016:
> @jared-2016 won't we also need to update each sample's .pro file to include the new PRI?

@ldanzinger I wasn't sure what our approach was going to be on this since there won't be a standardized location for the pri. I guess we can do something similar to what we do for the `ssl` files and if it does't locate it when it's needed it throws and error. Is that a reasonable option?

ldanzinger:
@jared-2016 I think we should consider doing 2 things:

* Add something to the sample readme that says how we recommend you setup the samples & toolkit repos - e.g. - clone the sample repo, cd into the local repo, clone the toolkit repo to the current directory, then we can assume that relative path for all individual samples. Similar to the "quick start" section of the DSA repo
* Add some qmake warnings that the developer must fix the path to the toolkit if it isn't set or isn't set correctly, just like how you mentioned above

ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMapLocalBasemap/GenerateOfflineMapLocalBasemap.qml
anmacdonald: There are more AuthenticationView declarations in this PR than I think we'd see in a typical application, so I'd say the benefit wouldn't be as effective in the wild.

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ShowPopup/ShowPopup.qml
anmacdonald: If we never got feedback from beta users on this, I would be tempted to say keep as is (but I do also have a personal investment in keeping the interface as pure/minimal as possible.) 

It is a minor addition so my arm can be twisted!

ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMapLocalBasemap/GenerateOfflineMapLocalBasemap.qml
JamesMBallard: I would prefer to avoid no-op properties. It would ease porting apps to the new toolkit, but we should wait to see if we hear feedback from users and consider adding it later.

JamesMBallard:
>I assume that we would like to remove all references to the departing version of the toolkit but didn't want to include that into this PR just incase.

Yes, and I would like to see that happen with this PR. We are making the jump so let's do it all the way.

ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMapLocalBasemap/GenerateOfflineMapLocalBasemap.qml
ldanzinger: yep I agree. Just wanted to throw it out there, as Qt is doing something similar with QList & QVector so as to make porting easier to Qt 6 https://www.qt.io/blog/qlist-changes-in-qt-6. In this case, this code is going to only be set once in an app, so it shouldn't be much to change

jared-2016:
@JamesMBallard Sounds good to me! I will get the rest of those in today.

ArcGISRuntimeSDKQt_QMLSamples/CloudAndPortal/ShowOrgBasemaps/ShowOrgBasemaps.qml
```diff
@@ -267,7 +267,7 @@ Rectangle {
     /*
     // Uncomment this section when running as standalone application
     AuthenticationView {
-        authenticationManager: AuthenticationManager
+        anchors.fill: parent
     }
     */
```
JamesMBallard, line 272: @jared-2016 can you confirm if this is still needed or not when running samples individually? This is currently commented out. There was a discussion at one point that these wouldn't be needed anymore, but I don't recall the specifics so I could be wrong on that.

JamesMBallard (approved):
Approving, but a few questions. No issues with the changes themselves.

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ShowPopup/ShowPopup.qml
ldanzinger: Let's leave it off and add it if we need it

ArcGISRuntimeSDKQt_QMLSamples/CloudAndPortal/ShowOrgBasemaps/ShowOrgBasemaps.qml
jared-2016: @JamesMBallard Yes the `AuthenticationView` is still needed for samples to be run individually.

There is another discussion about if the sample viewer apps would need one as well if they were defined in the individual samples but that is less clear as some samples which don't normally require one might need one in some rare scenarios. Lucas touched on this briefly in the sample viewer PR if you want to take a look there.

jared-2016:
Please forgive me as quite a few changes have come in with these last few commits.

1. Updating samples that utilize the toolkit to include the path to the .pri in their respective .pro files. Instructions to follow in readme so this is an expected place as mentioned by Lucas previously.
2. Removing references to AuthenticationManager.h and related boilerplate in .cpp/.h files that also include registering as a non create-able type, all of which are not necessary with the new toolkit.
3. Updating all main.cpp files to remove the old references to the toolkit(e.g. adding the toolkit to the import path etc.)

There was one exception for `OAuthRedirectHandler` cpp sample which after reviewing didn't seem to need any changes but please correct me if I am wrong on that one.

Please take another glance over these last changes to make sure nothing creeped in but these should be good moving forward.

PR #1134, opened 2020-11-02
updating URL and camera position

tanneryould:
updated cpp with @ldanzinger. Not ready for review, need to update QML and screenshot


ldanzinger (approved):
Code looks good

PR #1137, opened 2020-11-02
Updated elevation source to https

tanneryould:
Update elevation source from http to https.


ldanzinger (approved):
Looks good

jared-2016 (approved):
Looks good to me!

On a side note. I would venture to guess that the QML samples template needs updating as well since these all look like relatively new samples and I believe the elevation layer is auto generated when 3d is selected.

PR #1140, opened 2020-11-04
Add "distance" argument to Camera

sbiswas96:
Adding a single property to the `Camera` object so the zoom level of the C++ and QML samples match each other. Thanks for reviewing!


tanneryould (approved):
I confirmed this resolves the issue on my end as well.

PR #1142, opened 2020-11-04
Tan11389/readme changes through aug 4

tanneryould:
Please verify that I didn't accidentally add any invalid syntax. Also, I'm not sure if we want to include it in here, but I found a loose comparison and updated it to strict in one of the QML files.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/EditData/EditFeaturesWithFeatureLinkedAnnotation/README.md
+++ b/ArcGISRuntimeSDKQt_CppSamples/EditData/EditFeaturesWithFeatureLinkedAnnotation/README.md
@@ -35,13 +35,9 @@ Both expressions were defined by the data author in ArcGIS Pro using [the Arcade
 
 ## Offline data
 
-1. Download the data from [ArcGIS Online](https://arcgisruntime.maps.arcgis.com/home/item.html?id=74c0c9fa80f4498c9739cc42531e9948).
-2. Extract the contents of the downloaded zip file to disk.
-3. Open your command prompt and navigate to the folder where you extracted the contents of the data from step 1.
-4. Push the data into the scoped storage of the sample app:
-`<userhome>/ArcGIS/Runtime/Data/geodatabase/loudoun_anno.geodatabase`
-
-
+Link | Local Location
+---------|-------|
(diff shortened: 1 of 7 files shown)


ArcGISRuntimeSDKQt_CppSamples/EditData/EditFeaturesWithFeatureLinkedAnnotation/README.md
```diff
@@ -35,9 +35,13 @@ Both expressions were defined by the data author in ArcGIS Pro using [the Arcade
 
 ## Offline data
 
-Link | Local Location
----------|-------|
-|[Loudoun geodatabase](https://arcgisruntime.maps.arcgis.com/home/item.html?id=74c0c9fa80f4498c9739cc42531e9948)|`<userhome>/ArcGIS/Runtime/Data/geodatabase/loudoun_anno.geodatabase`|
+1. Download the data from [ArcGIS Online](https://arcgisruntime.maps.arcgis.com/home/item.html?id=74c0c9fa80f4498c9739cc42531e9948).
```
ldanzinger, line 38: I believe the offline data section is formatted a bit differently per SDK. In Qt, we use the table format. Also, the wording here with "scoped storage" is very Android specific - we try and keep it more platform agnostic in the Qt docs. I'd just revert this change (and the corresponding QML sample update) and stick with what we had

tanneryould: Ok, good to learn. I've gone ahead and reverted the changes.

jared-2016: I second what Lucas said here. Only thing you might consider adding is that on occasion we reference the samples readme for offline data, https://github.com/Esri/arcgis-runtime-samples-qt#use-offline-data-in-the-samples.

Here is an example, https://github.com/Esri/arcgis-runtime-samples-qt/tree/master/ArcGISRuntimeSDKQt_CppSamples/Scenes/Animate3DSymbols#offline-data.

ldanzinger (approved):
LGTM

jared-2016 (approved):
Just one opinion but either way, looks good!

PR #1149, opened 2020-11-13
Jared/toolkitfix

jared-2016:
@ldanzinger @anmacdonald Ready for review.

I combined a few other toolkit related fixes into this PR.

1. All Cpp samples that utilize the toolkit have been updated to pass in the engine argument
2. All Qml samples that utilize the toolkit have been updated to pass in the engine argument `*(view.engine())`.
3. TokenAuthentication screenshot has been updated

## AR related changes:
1. AR samples updated toolkit path in pro for all samples to reference the new path to the ar toolkit pri files.
2. Removed android related deployment files as this caused more problems moving forward than it solved
  - This was initially done for simplicity for the user to just pull the sample and be able to deploy but this caused problems with android manifest as it was referencing some items that were no longer there. So to avoid out of date manifest etc. The readme now instructs the user to follow the steps listed in the toolkit for deploying AR samples.
- since these files were being updated I also removed the deployment.pri files for the QML AR samples as they are no longer needed.
3. Updated Explore Scenes in Flyover AR to use the new integrated mesh layer and updated the screenshot's as well.


anmacdonald (approved):
LGTM

ArcGISRuntimeSDKQt_CppSamples/AR/DisplayScenesInTabletopAR/DisplayScenesInTabletopAR.pro
```diff
@@ -80,15 +81,6 @@ android {
...
     ANDROID_ABIS = arm64-v8a
-
-    DISTFILES += \
-        android/AndroidManifest.xml \
-        android/build.gradle \
-        android/gradle/wrapper/gradle-wrapper.jar \
-        android/gradle/wrapper/gradle-wrapper.properties \
-        android/gradlew \
-        android/gradlew.bat \
-        android/res/values/libs.xml
```
ldanzinger, line 91: Do you want to remove all this? I thought it was required still

ldanzinger (approved):
Just the one question - do we not need these for AR anymore for some reason?

ArcGISRuntimeSDKQt_CppSamples/AR/DisplayScenesInTabletopAR/DisplayScenesInTabletopAR.pro
jared-2016: @ldanzinger We do. The problem is that when we generate these templates at an earlier time it causes problems down the road. When I was testing my changes it was complaining about not being able to find some libraries that were specified in the manifest. This was similar to a problem I ran into with James previously and if I remember right it was referencing a deprecated library then and we didn't know why. The problem was solved in this case by just redoing the process specified in the AR toolkit. So I think this is more of a headache than it's worth since following the steps in the toolkit is really straight forward and it should avoid the problem I kept running into.

So in replacement of trying to do it for the end user, I just made reference to the toolkit instructions for deploying to android/iOS in the readme.

Should we continue to provide these upfront? or leave it up to the user?

PR #1161, opened 2020-12-01
Updating Change basemap, Display a Map, and Display a Scene to use basemap styles

jared-2016:


jared-2016:
accidentally opened before ready.

PR #1165, opened 2020-12-03
Remove redirect_from values check

tanneryould:
Metadata JSON files do not need to contain "redirect_from" values. This change modifies the check so it still ensures that "redirect_from" is included among the JSON names, but it does not verify that it contains values. This initial commit is a quick fix; if we want to expand on this, checks could ensure that if redirect values are listed, that they are valid URLs.


jared-2016 (approved):
LGTM. Not sure where they stands on the code freeze so hold off on merging unless @ldanzinger says otherwise.

PR #1169, opened 2020-12-10
Update source of packages to ArcGISRuntimeSDK account

tanneryould:
This updates the item ids of Local Server samples to point to our ArcGISRuntimeSDK account. Copies of the relevant mpkx/gpkx files have been uploaded there, categorized, and tagged.
@ldanzinger - should I delete the old items from the QtDev account?


ldanzinger (approved):
@tanneryould Looks good. Let's get this change checked in now. I want our sample viewers to ship with the proper portal items so we don't accidentally remove something important from our test orgs. You can delete the old items since we haven't released any software referencing them

PR #1171, opened 2021-01-11
updating the version number from 100.10 to 100.11

mfeigl:
Updating the version from 100.10 to 100.11


anmacdonald (approved):
LGTM, I'd ignore check, not your concern.

PR #1175, opened 2021-01-29
Cherry picking basemapStyle/API branch into master

tanneryould:


jared-2016 (approved):
Looks good to me. 👍

tanneryould:
The failing checks are a false negative and the result of a sort that is case sensitive and in the title check. I need to update those so the sort is case insensitive and the title check is a bit more lenient.

tanneryould:
@ldanzinger were you able to take a look at this?

ldanzinger (approved):
Spot checked QML & C++. Looks good

tanneryould:
Changes approved. Merging into master.

PR #1179, opened 2021-02-10
Android sample viewer find a place crash bug fix

tanneryould:
Resolves an issue where the Android sample viewer crashes when loading the "Find a place" sample and the user has not already allowed location permissions. Originally the sample would attempt to get the user's location immediately, but this update makes the sample wait until it has finished drawing the map view to get the user's location.


tanneryould:
@ldanzinger let me know who else I should bring in to review this change, thanks

ldanzinger (approved):
LGTM - one review is probably sufficient for this

PR #1180, opened 2021-02-10
Display utility TOC fix

tanneryould:
The display utility associations samples differed between C++ and QML (see below). This updates the colors in the QML sample to be consistent with those of the C++ sample.

b2877f00-38c1-11eb-9ea1-edbb471b4e7a


tanneryould:
The swatches were backwards. It looks like this is resolved for now.

Screen Shot 2021-02-10 at 10 49 00 AM

ldanzinger (approved):

PR #1186, opened 2021-02-19
Update basemap

tanneryould:
Fixes the OSM basemap style enum so the basemap properly displays.


tanneryould:
@ldanzinger - is this something that should be merged into master?

ldanzinger (approved):
let's get it fixed in master branch too

PR #1193, opened 2021-03-08
Fix corrupt PNG images

patrickarlt:
This fixes a build error with 2 corrupt PNG images.


tanneryould (approved):
LGTM - Works on my machine and displays in GitHub READMEs.

PR #1194, opened 2021-03-08
Tan11389/create symbol styles from web styles

tanneryould:
This creates a sample that demonstrates how to utilize a SymbolStyle from a web source.


ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/CreateSymbolStylesFromWebStyles/CreateSymbolStylesFromWebStyles.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include "pch.hpp"
+#endif // PCH_BUILD
+
+#include "CreateSymbolStylesFromWebStyles.h"
+
+#include "FeatureLayer.h"
+#include "ServiceFeatureTable.h"
+#include "Map.h"
+#include "MapQuickView.h"
+
```
ldanzinger, line 27:
```suggestion
```

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/CreateSymbolStylesFromWebStyles/CreateSymbolStylesFromWebStyles.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include "SymbolStyle.h"
+#include "SimpleMarkerSymbol.h"
+
+using namespace Esri::ArcGISRuntime;
+
+CreateSymbolStylesFromWebStyles::CreateSymbolStylesFromWebStyles(QObject* parent /* = nullptr */):
+  QObject(parent),
+  m_map(new Map(BasemapStyle::ArcGISNavigation, this))
+{
+  QUrl webStyleLayerUrl = QUrl("http://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/LA_County_Points_of_Interest/FeatureServer/0");
```
ldanzinger, line 38: should always use https

```suggestion
  QUrl webStyleLayerUrl = QUrl("https://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/LA_County_Points_of_Interest/FeatureServer/0");
```

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/CreateSymbolStylesFromWebStyles/CreateSymbolStylesFromWebStyles.cpp
```diff
@@ -0,0 +1,166 @@
...
+}
+
+void CreateSymbolStylesFromWebStyles::createSymbolStyles()
+{
+  m_categoriesMap = createCategoriesMap();
+  m_symbolsFetchedCount = 0;
+
+  for (const QString &symbolKey : m_categoriesMap.keys())
+  {
+    SymbolStyle* symbolStyle = new SymbolStyle("Esri2DPointSymbolsStyle", {}, this);
```
ldanzinger, line 72: I'm not sure that this is the optimal pattern to showcase. SymbolStyle is a container that has a bunch of symbols in it that you can fetch on demand. Instead of creating a new SymbolStyle instance every time you want a new symbol, wouldn't it make more sense to create one instance, and call fetchSymbol on that same instance for each key you want to fetch?

ArcGISRuntimeSDKQt_QMLSamples/DisplayInformation/CreateSymbolStylesFromWebStyles/CreateSymbolStylesFromWebStyles.qml
```diff
@@ -0,0 +1,227 @@
...
+                    // The UniqueValueRenderer will affect specific features based on the values of the specified FieldName(s)
+                    fieldNames: ["cat2"]
+                }
+
+                Component.onCompleted: {
+                    const symbolKeys = ["atm", "beach", "campground", "city-hall", "hospital", "library", "park", "place-of-worship", "police-station", "post-office", "school", "trail"];
+                    let symbolsFetchedCount = 0;
+
+                    symbolKeys.forEach((symbolKey) => {
+                                           var symbolStyle = ArcGISRuntimeEnvironment.createObject("SymbolStyle", {styleName: "Esri2DPointSymbolsStyle", portal: {}});
```
ldanzinger, line 58: As mentioned on the C++ sample above, I think it would be better to declare 1 SymbolStyle instance, setup 1 signal handler, and call fetchSymbol on that one instance for each key. One potential tricky part is that our QML API will only let you run one fetchSymbol at a time, so you will need to queue them up - C++ will allow you to run several concurrently.

ldanzinger (changes requested):

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/CreateSymbolStylesFromWebStyles/CreateSymbolStylesFromWebStyles.cpp
tanneryould: I thought that would be the case too, but it doesn't work if we use that pattern.

ldanzinger: > but it doesn't work if we use that pattern.

That doesn't seem right. What doesn't work?

ldanzinger: can you move your signal connection to be outside/before the for loop? right under the object instantiation.

tanneryould:
@ldanzinger I redesigned this to use `searchSymbols` instead of `fetchSymbols` and it's much cleaner now in regards to how we handle the legend. `searchSymbols` returns a QAbstractList that we can just plug into the `ListView` delegate. The list is created in alphabetical order and contains no duplicates.

I don't love having to make nested connections, so if that's not the ideal code paradigm please let me know.

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/CreateSymbolStylesFromWebStyles/CreateSymbolStylesFromWebStyles.cpp
```diff
@@ -0,0 +1,149 @@
...
+  m_map->setReferenceScale(100'000);
+
+  createSymbolStyles();
+}
+
+void CreateSymbolStylesFromWebStyles::createSymbolStyles()
+{
+  QMap<QString,QStringList> categoriesMap = createCategoriesMap();
+  SymbolStyle* symbolStyle = new SymbolStyle("Esri2DPointSymbolsStyle", {}, this);
+  SymbolStyleSearchParameters searchParams = SymbolStyleSearchParameters();
```
ldanzinger, line 62:
```suggestion
  SymbolStyleSearchParameters searchParams;
```

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/CreateSymbolStylesFromWebStyles/CreateSymbolStylesFromWebStyles.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+void CreateSymbolStylesFromWebStyles::setMapView(MapQuickView* mapView)
+{
+  if (!mapView || mapView == m_mapView)
+    return;
+
+  m_mapView = mapView;
+  m_mapView->setMap(m_map);
+
+  const double x_longitude = -118.44186;
```
ldanzinger, line 117: doubles that are known at compile time can be marked as constexpr

ArcGISRuntimeSDKQt_QMLSamples/DisplayInformation/CreateSymbolStylesFromWebStyles/CreateSymbolStylesFromWebStyles.qml
```diff
@@ -0,0 +1,201 @@
...
+                    y: 34.28301
+                    spatialReference: SpatialReference { wkid: 4326 }
+                }
+                targetScale: 7000
+            }
+        }
+        onMapScaleChanged: {
+            // Set scale symbols to true when we zoom in so the symbols don't take up the entire view
+            webLayer.scaleSymbols = (mapView.mapScale >= 80000);
+        }
```
ldanzinger, line 97: would be better to use property binding instead of assignment if possible. 

for example

```
FeatureLayer {
    id: webLayer
    scaleSymbols: mapView.mapScale >= 8000
    ServiceFeatureTable {
        url: "https://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/LA_County_Points_of_Interest/FeatureServer/0"
    }
```

ArcGISRuntimeSDKQt_QMLSamples/DisplayInformation/CreateSymbolStylesFromWebStyles/CreateSymbolStylesFromWebStyles.qml
```diff
@@ -0,0 +1,201 @@
...
+                                                                                                                                                                                        const uniqueValue = ArcGISRuntimeEnvironment.createObject("UniqueValue", {label: label, values: [value], symbol: symbol}, webLayer);
+                                                                                                                                                                                        uniqueValueRenderer.uniqueValues.append(uniqueValue);
+                                                                                                                                                                                    });
+                                                                                                                                                                 });
+                                                                                                                   symbolResult.fetchSymbol();
+                                                                                                               });
```
ldanzinger, line 77: This is pretty gnarly to read...

In QML connect statements, you can also just pass in a JS function. It might be better to do that than to have several nested arrow functions.

We do this in other samples, for example here we are connecting to a JS function that is created a few lines below it. It can help keep things a bit more tidy and readable. https://github.com/Esri/arcgis-runtime-samples-qt/blob/master/ArcGISRuntimeSDKQt_QMLSamples/Maps/GenerateOfflineMap/GenerateOfflineMap.qml#L116

ldanzinger (changes requested):
This looks a lot better. Couple more changes needed

tanneryould:
@ldanzinger I addressed your comments and added functions to the QML connections. Couple of questions:
1. What do you think of the names I used for the functions `searchSymbolHandler()` and `fetchSymbolHandler()` respectively? 
2. Should `symbolStyle` and `symbolStyleSearchParameters` be declared in the `webLayer` object? I tried that, but when I went to retrieve `symbolStyle` in `searchSymbolHandler()`, it returned `undefined`. Keeping the two variables as they were was the path of least resistance in addressing the code changes.

ArcGISRuntimeSDKQt_QMLSamples/DisplayInformation/CreateSymbolStylesFromWebStyles/CreateSymbolStylesFromWebStyles.qml
```diff
@@ -0,0 +1,207 @@
...
+                }
+
+                Component.onCompleted: {
+                    const symbolStyle = ArcGISRuntimeEnvironment.createObject("SymbolStyle", {styleName: "Esri2DPointSymbolsStyle"}, webLayer);
+
+                    const symbolStyleSearchParameters = ArcGISRuntimeEnvironment.createObject("SymbolStyleSearchParameters");
+                    symbolStyleSearchParameters.keys = ["atm", "beach", "campground", "city-hall", "hospital", "library", "park", "place-of-worship", "police-station", "post-office", "school", "trail"];
+                    symbolStyleSearchParameters.keysStrictlyMatch = true;
+
+                    symbolStyle.searchSymbolsStatusChanged.connect(() => searchSymbolsHandler(symbolStyle));
```
ldanzinger, line 62: I think you can do something like this:
```suggestion
                    symbolStyle.searchSymbolsStatusChanged.connect(searchSymbolsHandler(symbolStyle));
```

ArcGISRuntimeSDKQt_QMLSamples/DisplayInformation/CreateSymbolStylesFromWebStyles/CreateSymbolStylesFromWebStyles.qml
```diff
@@ -0,0 +1,207 @@
...
+                    symbolStyle.searchSymbols(symbolStyleSearchParameters);
+                }
+
+                function searchSymbolsHandler(symbolStyle) {
+                    if (symbolStyle.searchSymbolsStatus !== Enums.TaskStatusCompleted)
+                        return;
+
+                    legendItems = symbolStyle.searchSymbolsResult;
+                    symbolStyle.searchSymbolsResult.forEach((symbolResult) => {
+                                                                symbolResult.fetchSymbolStatusChanged.connect(() => fetchSymbolsHandler(symbolResult));
```
ldanzinger, line 72: I think you can do something like this:

```suggestion
                                                                symbolResult.fetchSymbolStatusChanged.connect(fetchSymbolsHandler(symbolResult));
```

ldanzinger (commented):
> What do you think of the names

Names look ok to me.

> Should symbolStyle and symbolStyleSearchParameters be declared in the webLayer object?

I think I would declare SymbolStyle at the global level (outermost level). You should be able to access that object from anywhere within that file.

```
WebStyle {
    styleName: ""
    Component.onCompleted {
        // execute your search here
    }
}
```

ldanzinger (approved):

tanneryould:
Adding create symbol styles from web styles sample

PR #1195, opened 2021-03-08
Tan11389/display 3d labels

tanneryould:
Creates a new sample to demonstrate how to add labels to a layer on a web scene.


ArcGISRuntimeSDKQt_CppSamples/Scenes/Display3DLabelsInScene/Display3DLabelsInScene.cpp
```diff
@@ -0,0 +1,110 @@
...
+#ifdef PCH_BUILD
+#include "pch.hpp"
+#endif // PCH_BUILD
+
+#include "Display3DLabelsInScene.h"
+
+#include "ArcGISTiledElevationSource.h"
+#include "Scene.h"
+#include "SceneQuickView.h"
+
```
ldanzinger, line 26:
```suggestion
```

ArcGISRuntimeSDKQt_CppSamples/Scenes/Display3DLabelsInScene/Display3DLabelsInScene.cpp
```diff
@@ -0,0 +1,110 @@
...
+        QUrl("https://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer"), this);
+
+  // add the elevation source to the scene to display elevation
+  m_scene->baseSurface()->elevationSources()->append(elevationSource);
+
+  connect(m_scene, &Scene::doneLoading, this, [this]()
+  {
+    for (Layer* layer : *m_scene->operationalLayers())
+    {
+      if(layer->name() == "Gas")
```
ldanzinger, line 51:
```suggestion
      if (layer->name() == "Gas")
```

ArcGISRuntimeSDKQt_CppSamples/Scenes/Display3DLabelsInScene/Display3DLabelsInScene.cpp
```diff
@@ -0,0 +1,110 @@
...
+        break;
+      }
+    }
+  });
+}
+
+void Display3DLabelsInScene::display3DLabelsOnFeatureLayer(FeatureLayer* featureLayer)
+{
+  TextSymbol* textSymbol = new TextSymbol(this);
+  textSymbol->setColor(QColor("#ffa500"));
```
ldanzinger, line 69: I was using this sample as a starting point for a demo at dev summit, and I received the following feedback. It might be good to consider if we update the sample data or text properties to make it look a bit sharper:

- text is a bit too small
- text color doesn't have good contrast with the layers in the map (yellow on white)
- what is the purpose of the building layer in the scene? is this an important part of the story of the data on this map and why we want to display these specific labels?

ArcGISRuntimeSDKQt_QMLSamples/Scenes/Display3DLabelsInScene/Display3DLabelsInScene.qml
```diff
@@ -0,0 +1,71 @@
...
+
+            onLoadStatusChanged: {
+                if (loadStatus !== Enums.LoadStatusLoaded)
+                    return;
+
+                // Search for a specific layer from the web scene's operational layers to apply labels to
+                nycScene.operationalLayers.forEach(layer => {
+                                                       if (layer.name === "Gas") {
+                                                           // The Gas layer is a GroupLayer type and we must extract our target FeatureLayer from it
+                                                           let gasLayer = layer.layers.get(0);
```
ldanzinger, line 61: const is preferred. In JS, const is still mutable, it just can't be reassigned. That should be the first variable type we try to use, then let, then var.

ldanzinger (changes requested):

ArcGISRuntimeSDKQt_CppSamples/Scenes/Display3DLabelsInScene/Display3DLabelsInScene.cpp
```diff
@@ -0,0 +1,109 @@
...
+  connect(m_scene, &Scene::doneLoading, this, [this]()
+  {
+    for (Layer* layer : *m_scene->operationalLayers())
+    {
+      if (layer->name() == "Gas")
+      {
+        // The gas layer is a GroupLayer type consisting of Layer types.
+        // Labels can only be displayed on FeatureLayer types, so we must first convert it to a FeatureLayer class.
+        GroupLayer* gasGroupLayer = dynamic_cast<GroupLayer*>(layer);
+        FeatureLayer* gasFeatureLayer = dynamic_cast<FeatureLayer*>(gasGroupLayer->layers()->first());
```
ldanzinger, line 55: I guess technically you should check if the cast for group layer worked before accessing its layers. That could return nullptr if the layer name is Gas but it cannot be casted to a GroupLayer.

```
        GroupLayer* gasGroupLayer = dynamic_cast<GroupLayer*>(layer);
        if (!gasGroupLayer)
          break; // or return/continue - whatever you think...
```

ldanzinger (approved):
Looks good. Marking as approved, but should add a null check for safety on the group layer cast. No need for re-approval before merging

PR #1205, opened 2021-03-31
Tan11389/blend raster sample zero fix

tanneryould:
* Resolves a bug where typed values would revert to zero after the user clicks away. The SpinBox errored when converting text with a degree symbol back into a value. This PR adds an additional function to parse the string into an int.
* Resolves a bug in QML where the basemap changes when the user selects a new color ramp from the ComboBox rather than when the user presses the "Render" button.
* Fixes some formatting


tanneryould:
Looking at the history it looks like azimuth was accidentally capped at 90 when the sliders were changed to SpinBoxes. I reset it to a full 360º. I also renamed the SpinBox elements because they are no longer sliders.

ldanzinger (approved):

PR #1206, opened 2021-04-01
Tan11389/scene animate img fix

tanneryould:
Fixes a bug where infinite `ImageFrame` objects were being created in the QML sample. Resolves this by adding the objects to a list and then referencing the list. Once the length of the list is equal to the length of the file list, it stops creating new objects and continues to loop through the completed list.

I do this as each `ImageFrame` is created rather than creating the list upon initialization of the `FileFolder` because the latter delays the app opening for about 3 to 4 seconds (on my machine) while loading the list. Creating the list as each image is created allows for a more seamless user experience.

I verified that the C++ sample is not affected by this logic bug.

### DIFF
--- a/ArcGISRuntimeSDKQt_QMLSamples/Scenes/AnimateImagesWithImageOverlay/AnimateImagesWithImageOverlay.qml
+++ b/ArcGISRuntimeSDKQt_QMLSamples/Scenes/AnimateImagesWithImageOverlay/AnimateImagesWithImageOverlay.qml
@@ -215,8 +215,6 @@ Rectangle {
         id: imageFrameFolder
         url: dataPath
 
-        Component.onCompleted: {
-            fileNamesLength = imageFrameFolder.fileNames().length;
-        }
+        Component.onCompleted: fileNamesLength = imageFrameFolder.fileNames().length;
     }
 }
--- a/ArcGISRuntimeSDKQt_QMLSamples/Scenes/AnimateImagesWithImageOverlay/AnimateImagesWithImageOverlay.qml
+++ b/ArcGISRuntimeSDKQt_QMLSamples/Scenes/AnimateImagesWithImageOverlay/AnimateImagesWithImageOverlay.qml
@@ -55,7 +55,7 @@ Rectangle {
             }
 
             // set image frame to image overlay
-            imageOverlay.imageFrame = imageFrameList[index]
+            imageOverlay.imageFrame = imageFrameList[index];
 
             // increment the index to keep track of which image to load next
             index++;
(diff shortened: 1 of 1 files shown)


ArcGISRuntimeSDKQt_QMLSamples/Scenes/AnimateImagesWithImageOverlay/AnimateImagesWithImageOverlay.qml
```diff
@@ -44,14 +45,17 @@ Rectangle {
...
-                                                                       extent: imageFrameExtent
-                                                                   });
+            // Create new ImageFrames from images in the folder
+            // Add them to a list so infinite ImageFrame objects are not created each loop
+            if (imageFrameList.length !== fileNamesLength) {
+                imageFrameList.push(ArcGISRuntimeEnvironment.createObject("ImageFrame", {
+                                                                              url: dataPath + "/" + imageFrameFolder.fileNames()[index],
+                                                                              extent: imageFrameExtent
+                                                                          }));
+            }
```
ldanzinger, line 55: Does this perform ok? No lag the first time? It might be better to put this in the Scene's loadStatusChanged signal handler once it is loaded. Then, once everything is ready to go, enable the button so users can begin the loop. This is just an idea - if everything works well, then this should work just fine too

ArcGISRuntimeSDKQt_QMLSamples/Scenes/AnimateImagesWithImageOverlay/AnimateImagesWithImageOverlay.qml
```diff
@@ -44,14 +45,17 @@ Rectangle {
...
 
             // set image frame to image overlay
-            imageOverlay.imageFrame = imageFrame;
+            imageOverlay.imageFrame = imageFrameList[index]
```
ldanzinger, line 58:
```suggestion
            imageOverlay.imageFrame = imageFrameList[index];
```

ldanzinger (approved):
Looks good, couple of comments to consider

ArcGISRuntimeSDKQt_QMLSamples/Scenes/AnimateImagesWithImageOverlay/AnimateImagesWithImageOverlay.qml
tanneryould: @ldanzinger There's no discernible difference between the initial loop and latter ones. The only difference between the previous version and the this PR is that we're pushing new objects to a list before referencing. I agree it's a bit of an odd workflow however.

PR #1209, opened 2021-04-15
updating the version number from 100.11 to 100.12

mfeigl:
Updating the version from 100.11 to 100.12


JamesMBallard (approved):
👍

JamesMBallard:
@tanneryould the git checks are failing for this. Do we need to address that before merging?

tanneryould:
@JamesMBallard fortunately there are only two errors. I'd like to resolve this before merging and it shouldn't take too long.

tanneryould:
@JamesMBallard ok, the check is raising a false error. The OAuth Redirect Example doesn't follow the standard md pattern so I'm going to open a PR to add this file as an extension. In the meantime, this is good to go.

tanneryould (approved):
Check raises false failure. All tests pass.

PR #1214, opened 2021-04-23
Removed default lambda captures from samples

MicahSamaduroff:
Lambda captures have been removed from the C++ and Widgets samples. These were removed as they can capture more data than necessary, and default captures of 'this' may be deprecated in the future.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Geometry/NearestVertex/NearestVertex.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Geometry/NearestVertex/NearestVertex.cpp
@@ -119,7 +119,9 @@ void NearestVertex::setupGraphics()
   graphicsOverlay->graphics()->append(nearestVertexGraphic);
 
   // add graphic to clicked location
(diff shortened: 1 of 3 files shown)


ArcGISRuntimeSDKQt_CppSamples/Geometry/NearestVertex/NearestVertex.cpp
```diff
@@ -119,7 +119,7 @@ void NearestVertex::setupGraphics()
   graphicsOverlay->graphics()->append(nearestVertexGraphic);
 
   // add graphic to clicked location
-  connect(m_mapView, &MapQuickView::mouseClicked, this, [=](QMouseEvent& e)
+  connect(m_mapView, &MapQuickView::mouseClicked, this, [nearestVertexGraphic, nearestCoordinateGraphic, polygonBuilder, clickedLocationGraphic, this](QMouseEvent& e)
```
JamesMBallard, line 122: Please wrap this over a few lines so it's not so long.

JamesMBallard (changes requested):

ArcGISRuntimeSDKQt_CppSamples/Geometry/NearestVertex/NearestVertex.cpp
MicahSamaduroff: This is done

JamesMBallard (approved):
👍

PR #1218, opened 2021-04-30
Fix KML callout

tanneryould:
## Background
In U11, core introduced a condition to return an empty string for the balloons/KML nodes that doesn't have description or extended data. The data source in the Identify KML features doesn't provide a description for each significant weather condition, thus empty balloons are displayed in the callout.

Before: the popup for any feature in the map is empty
After: the popup should show HTML content

## Fix
If description is absent for a placemark, add a placeholder so that the content of the balloon can be displayed.


tanneryould:
I changed the width from 400 to 300 because, for whatever reason, if the callout is only one line there's a blue bar that appears below it.

ldanzinger (approved):

PR #1226, opened 2021-05-25
Tan11389/create load report qml fix

tanneryould:
There's an issue with the way the list of QmlCodedValue when retrieved by `utilityNetwork.definition.networkAttribute(phasesNetworkAttributeName).domain.codedValues`. After repeated calls to get the attributes of the QmlCodedValue list, a Seg Fault was raised and the sample viewer crashed.

This fix constructs an array of coded values instead. This seems to be more stable and the sample viewer does not crash after repeated calls to the coded value array.


tanneryould:
For some reason this is still showing changes previously made to update the sample?

ldanzinger (approved):

PR #1230, opened 2021-05-27
Tan11389/locator update

tanneryould:
Updates samples to use new geocoding services and offline locators. Tested on individual samples and sample viewer on mac. Tested in sample viewer on Windows. It all runs as expected.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Search/FindAddress/FindAddress.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Search/FindAddress/FindAddress.cpp
@@ -78,6 +78,7 @@ void FindAddress::componentComplete()
 
   // create locator task and parameters
   //! [FindAddress create LocatorTask]
+  // An ArcGIS Developer API key is required to utilize this world geocoding service
   m_locatorTask = new LocatorTask(QUrl("https://geocode-api.arcgis.com/arcgis/rest/services/World/GeocodeServer"), this);
   //! [FindAddress create LocatorTask]
   m_geocodeParameters.setMinScore(75);
--- a/ArcGISRuntimeSDKQt_CppSamples/Search/FindAddress/FindAddress.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Search/FindAddress/FindAddress.cpp
@@ -78,7 +78,7 @@ void FindAddress::componentComplete()
 
   // create locator task and parameters
   //! [FindAddress create LocatorTask]
-  m_locatorTask = new LocatorTask(QUrl("https://geocode.arcgis.com/arcgis/rest/services/World/GeocodeServer"), this);
+  m_locatorTask = new LocatorTask(QUrl("https://geocode-api.arcgis.com/arcgis/rest/services/World/GeocodeServer"), this);
   //! [FindAddress create LocatorTask]
   m_geocodeParameters.setMinScore(75);
   m_geocodeParameters.setResultAttributeNames(QStringList { "Place_addr", "Match_addr" });
--- a/ArcGISRuntimeSDKQt_CppSamples/Search/OfflineGeocode/README.md
+++ b/ArcGISRuntimeSDKQt_CppSamples/Search/OfflineGeocode/README.md
@@ -31,7 +31,7 @@ Read more about how to set up the sample's offline data [here](http://links.esri
 Link | Local Location
 ---------|-------|
 |[SanDiego tpkx File](https://www.arcgis.com/home/item.html?id=22c3083d4fa74e3e9b25adfc9f8c0496)| `<userhome>`/ArcGIS/Runtime/Data/tpkx/streetmap_SD.tpkx |
-|[SanDiego loc Files](https://www.arcgis.com/home/item.html?id=c88c170e18f74740a0e4c660a5ff51c4)| `<userhome>`/ArcGIS/Runtime/Data/Locators/SanDiegoStreetAddress/SanDiego_StreetAddress.loc |
+|[SanDiego loc Files](https://www.arcgis.com/home/item.html?id=3424d442ebe54f3cbf34462382d3aebe)| `<userhome>`/ArcGIS/Runtime/Data/Locators/SanDiegoStreetAddress/SanDiego_StreetAddress.loc |
 
 ## Tags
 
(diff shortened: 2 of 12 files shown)


ArcGISRuntimeSDKQt_CppSamples/Search/FindAddress/FindAddress.cpp
```diff
@@ -78,7 +78,7 @@ void FindAddress::componentComplete()
 
   // create locator task and parameters
   //! [FindAddress create LocatorTask]
-  m_locatorTask = new LocatorTask(QUrl("https://geocode.arcgis.com/arcgis/rest/services/World/GeocodeServer"), this);
+  m_locatorTask = new LocatorTask(QUrl("https://geocode-api.arcgis.com/arcgis/rest/services/World/GeocodeServer"), this);
```
ldanzinger, line 81: optional - might be worthwhile to add a comment indicating that api keys will be required

ArcGISRuntimeSDKQt_CppSamples/Search/OfflineGeocode/README.metadata.json
```diff
@@ -6,20 +6,8 @@
             "path": "~/ArcGIS/Runtime/Data/tpkx/streetmap_SD.tpkx"
         },
         {
-            "itemId": "c88c170e18f74740a0e4c660a5ff51c4",
+            "itemId": "3424d442ebe54f3cbf34462382d3aebe",
             "path": "~/ArcGIS/Runtime/Data/Locators/SanDiegoStreetAddress/SanDiego_StreetAddress.loc"
```
ldanzinger, line 10: This might be worth testing out the following scenario:

* run the old viewer and download the data for offline geocode sample
* run the new viewer with these changes, and open offline geocode - it will find the old locator and not prompt to download the new one

This means that samples will continue to run against the older data if it has already been downloaded. It might be worth changing the file path slightly so that the new locators go to a new location. This will ensure the sample viewer downloads the new data, while old viewers can continue running against the old data

ArcGISRuntimeSDKQt_QMLSamples/Search/OfflineGeocode/README.metadata.json
```diff
@@ -6,8 +6,8 @@
             "path": "~/ArcGIS/Runtime/Data/tpkx/streetmap_SD.tpkx"
         },
         {
-            "itemId": "c88c170e18f74740a0e4c660a5ff51c4",
-            "path": "~/ArcGIS/Runtime/Data/Locators/SanDiegoStreetAddress/SanDiego_StreetAddress.zip"
```
ldanzinger, line 10: did you test this out? I seem to remember some nuance in the sample viewer download codepath where if you download a zip file, it will unzip it automatically, whereas if you specify the individual file within the zip file, it might not behave as expected. I can't remember all the details exactly...

tanneryould: @ldanzinger the C++ and QML implementations of Offline geocode differed in this. C++ looked for the `.loc` file and QML looked for the `.zip` file. Both workflows work, so in order to standardize the two, I went with the C++ workflow. If you feel that getting the `.zip` would be more stable we can opt for that instead.

ArcGISRuntimeSDKQt_CppSamples/Search/OfflineGeocode/README.metadata.json
tanneryould: Good point. The 100.12 implementation of the offline geocode in the sample viewer doesn't work if the old locator is downloaded to that location. Any thoughts on what the new path should be? I'm thinking `"~/ArcGIS/Runtime/Data/Locators/SanDiegoStreetAddressLocator/SanDiego_StreetAddress.loc"`. Appending "Pro" to the end instead wouldn't make sense because they were both made in Pro. The name for these new locators is the "eagle locator" but if we put that in, people may be misled if they expect a map with eagle locations on it.

ldanzinger: Your suggestion works for me

ArcGISRuntimeSDKQt_CppSamples/Search/OfflineGeocode/README.md
```diff
@@ -31,7 +31,7 @@ Read more about how to set up the sample's offline data [here](http://links.esri
 Link | Local Location
 ---------|-------|
 |[SanDiego tpkx File](https://www.arcgis.com/home/item.html?id=22c3083d4fa74e3e9b25adfc9f8c0496)| `<userhome>`/ArcGIS/Runtime/Data/tpkx/streetmap_SD.tpkx |
-|[SanDiego loc Files](https://www.arcgis.com/home/item.html?id=c88c170e18f74740a0e4c660a5ff51c4)| `<userhome>`/ArcGIS/Runtime/Data/Locators/SanDiegoStreetAddress/SanDiego_StreetAddress.loc |
+|[SanDiego loc Files](https://www.arcgis.com/home/item.html?id=3424d442ebe54f3cbf34462382d3aebe)| `<userhome>`/ArcGIS/Runtime/Data/Locators/SanDiegoStreetAddress/SanDiego_StreetAddress.loc |
```
ldanzinger, line 34: new path needs to be updated here too

ArcGISRuntimeSDKQt_QMLSamples/Search/OfflineGeocode/README.md
```diff
@@ -31,7 +31,7 @@ Read more about how to set up the sample's offline data [here](http://links.esri
 Link | Local Location
 ---------|-------|
 |[SanDiego tpkx file](https://www.arcgis.com/home/item.html?id=22c3083d4fa74e3e9b25adfc9f8c0496)| `<userhome>`/ArcGIS/Runtime/Data/tpkx/streetmap_SD.tpkx |
-|[SanDiego loc Files](https://www.arcgis.com/home/item.html?id=c88c170e18f74740a0e4c660a5ff51c4)| `<userhome>`/ArcGIS/Runtime/Data/Locators/SanDiegoStreetAddress/SanDiego_StreetAddress.loc |
+|[SanDiego loc Files](https://www.arcgis.com/home/item.html?id=3424d442ebe54f3cbf34462382d3aebe)| `<userhome>`/ArcGIS/Runtime/Data/Locators/SanDiegoStreetAddress/SanDiego_StreetAddress.loc |
```
ldanzinger, line 34: here too

tanneryould: Good catch thank you

ldanzinger (approved):

PR #1232, opened 2021-06-08
Fix animate images sample

MicahSamaduroff:
The sample viewer had high memory usage after running the sample "Animate Images with Overlay". This is because the `imageFrameList` is declared as a property, and was QML was not running the garbage collector on it until we restarted the sample.

This PR addresses this by freeing memory used by the image frames when the sample is navigated away from.


JamesMBallard (approved):
Nice find @MicahSamaduroff.

PR #1233, opened 2021-06-08
Updated how imageFrameList is cleared

MicahSamaduroff:
Updated removing all elements from `imageFrameList` from using a while loop with `pop()` to just clearing it. 

This makes it more clear what is happening, and also might be faster.


tanneryould (approved):
LGTM thanks!

JamesMBallard (approved):
Even better

PR #1236, opened 2021-06-15
Update unique pointers to C++ 14 standard

tanneryould:
I hadn't pushed these changes up, so this updates `m_tempDir` to use `std::make_unique`


ldanzinger (approved):
good cleanup

PR #1240, opened 2021-06-21
Tan11389/export tiles fix

tanneryould:
~~Sample server 6 does not support exporting to `.tpkx`, so this PR reverts the sample to exporting to `.tpk` instead. This is consistent with the other SDKs - all of which still export to `.tpk`s.~~

The PR updates the service url to pull tiles from the ArcGIS Imagery BasemapStyle raster base layer. Because of the change in service, we are limited by how many tiles we can export, so I have adjusted the scope of the export as well.


ldanzinger (approved):

ArcGISRuntimeSDKQt_CppSamples/Layers/ExportTiles/ExportTiles.cpp
```diff
@@ -56,22 +56,33 @@ void ExportTiles::componentComplete()
...
 
   // create a tiled basemap
-  ArcGISTiledLayer* tiledLayer = new ArcGISTiledLayer(m_serviceUrl, this);
-  Basemap* basemap = new Basemap(tiledLayer, this);
+  Basemap* basemap = new Basemap(BasemapStyle::ArcGISImagery, this);
+
+  // create an export tile cache task when basemap has finished loading
+  connect(basemap, &Basemap::doneLoading, this, [this]()
+  {
+    if (m_map->basemap()->baseLayers()->size() > 0)
```
ldanzinger, line 64: I suggest using isEmpty instead - one less comparison to make - https://developers.arcgis.com/qt/cpp/api-reference/esri-arcgisruntime-layerlistmodel.html#isEmpty

ArcGISRuntimeSDKQt_CppSamples/Layers/ExportTiles/ExportTiles.cpp
```diff
@@ -163,7 +173,10 @@ void ExportTiles::displayOutputTileCache(TileCache* tileCache)
   {
     if (tiledLayer->loadStatus() == LoadStatus::Loaded)
     {
-      m_mapView->setViewpointScale(m_mapView->mapScale() * 0.5);
+      double prevMapScale = m_mapView->mapScale();
```
ldanzinger, line 176: const

PR #1242, opened 2021-06-25
Enable the new RouteTracker callback functionality.

JamesMBallard:
This optimizes the voice guidance so it knows when the TTS
engine is ready.

@tanneryould please review. Aside from the changes mentioned I make some minor updates to the NavigateRouteSpeaker class, as well as making it a declarable object and no longer a singleton.


tanneryould (approved):
LGTM; thanks for taking this on

PR #1246, opened 2021-07-06
Fix build error

JamesMBallard:
@tanneryould please review. I found that I'm getting a build error due to this on a non-PCH build config on macOS.


tanneryould (approved):
LGTM

PR #1247, opened 2021-07-07
Renamed metadata titles for consistency between C++ and Qml.

AndrewBladon:
Fixing inconsistencies between sample names in C++ and QML.


tanneryould (approved):
LGTM

PR #1249, opened 2021-07-07
Tan11389/100.11.2/stage

tanneryould:
Merges changes from the 100.11.2 sample viewer release into main. Staged from a different release branch to avoid any issue with the sample viewer rebuild.


tanneryould:
I'm not sure why we're having so many merge conflicts and it's bringing in all of these old changes, but after taking a look through the files, this looks to be the full update from 100.11.0 to 100.11.2

PR #1270, opened 2021-08-09
Fix sbbg url

tanneryould:
Fixes the URL to the Santa Barbara Botanic Garden website by adding `www.`


AndrewBladon (approved):
After confirming the previous url does not work, the updated version (including "www.") works as expected.

PR #1272, opened 2021-08-09
Tan11389/sketch delete button fix

tanneryould:
The sketch on map sample would crash if "delete selected vertex" was called on an empty sketch.
The delete button popup also displayed incorrectly in the Sample Viewer - this change sets the width to allow the box to resize to fit the inner text.


AndrewBladon (approved):
Look good to me @tanneryould!

PR #1273, opened 2021-08-10
Andrewbladon/fix ogc feature service crash

AndrewBladon:
This PR is to fix crashes found during verification. Investigation into the cause suggested that if the load buttons were pressed while the feature service or a feature layer was loading, the sample viewer crashed. To prevent this, the load buttons have been disabled while the feature service or a feature layer is loading.

@ldanzinger and @JamesMBallard, could you confirm that, once approved, it is ok to merge these changes into the 100.12.0 release branch?

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.cpp
@@ -99,9 +99,7 @@ void BrowseOGCAPIFeatureService::handleError(const Esri::ArcGISRuntime::Error& e
 void BrowseOGCAPIFeatureService::loadFeatureService(const QUrl& url)
 {
   if (m_featureService && m_featureService->loadStatus() == LoadStatus::Loading)
-  {
     return;
-  }
 
   clearExistingFeatureService();
 
@@ -114,8 +112,10 @@ void BrowseOGCAPIFeatureService::loadFeatureService(const QUrl& url)
   // Connect errorOccurred to handleError()
   connect(m_featureService, &OgcFeatureService::errorOccurred, this, &BrowseOGCAPIFeatureService::handleError);
 
+  // Connect to loadingChanged() to enable and disable UI buttons
+  connect(m_featureService, &OgcFeatureService::loadStatusChanged, this, &BrowseOGCAPIFeatureService::loadingChanged);
+
   m_featureService->load();
-  setLoading(true);
 }
 
 void BrowseOGCAPIFeatureService::clearExistingFeatureService()
@@ -131,14 +131,13 @@ void BrowseOGCAPIFeatureService::clearExistingFeatureService()
     delete m_featureService;
     m_featureService = nullptr;
   }
+
   m_featureCollectionList.clear();
   emit featureCollectionListChanged();
 }
 
 void BrowseOGCAPIFeatureService::retrieveCollectionInfos()
 {
-  setLoading(false);
-
   // Assign OGC service metadata to m_serviceInfo property
   m_serviceInfo = m_featureService->serviceInfo();
 
@@ -196,7 +195,10 @@ void BrowseOGCAPIFeatureService::loadFeatureCollection(int selectedFeature)
   // Connect errorOccurred to handleError()
   connect(m_featureLayer, &FeatureLayer::errorOccurred, this, &BrowseOGCAPIFeatureService::handleError);
 
-  setLoading(true);
+  // Connect to loadingChanged() to enable and disable UI buttons
+  connect(m_featureLayer, &FeatureLayer::loadStatusChanged, this, &BrowseOGCAPIFeatureService::loadingChanged);
+
+  m_featureLayer->load();
 }
 
 void BrowseOGCAPIFeatureService::clearExistingFeatureLayer()
@@ -216,8 +218,6 @@ void BrowseOGCAPIFeatureService::clearExistingFeatureLayer()
 
 void BrowseOGCAPIFeatureService::addFeatureLayerToMap()
 {
-  setLoading(false);
-
   // Adjust the viewpoint to match the extent of the layer
   m_mapView->setViewpointGeometry(m_featureLayer->fullExtent());
 
@@ -228,11 +228,11 @@ void BrowseOGCAPIFeatureService::addFeatureLayerToMap()
 
 bool BrowseOGCAPIFeatureService::loading() const
 {
-  return isLoading;
-}
+  if (m_featureService && m_featureService->loadStatus() == LoadStatus::Loading)
+    return true;
 
-void BrowseOGCAPIFeatureService::setLoading(bool loading)
-{
-  isLoading = loading;
-  emit loadingChanged();
+  else if (m_featureLayer && m_featureLayer->loadStatus() == LoadStatus::Loading)
+    return true;
+
+  return false;
 }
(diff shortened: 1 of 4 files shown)


ArcGISRuntimeSDKQt_QMLSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.qml
```diff
@@ -124,7 +127,15 @@ Rectangle {
         // Connect loaded signal to checkForLoadingErrors() function
         root.featureService.loadStatusChanged.connect(checkForLoadingErrors);
 
+        // Once loaded, return the loading property to false
+        root.featureService.loadStatusChanged.connect(function() {
+            if (root.featureService.loadStatus === Enums.LoadStatusLoaded)
+                loading = false;
```
lsmallwood, line 133: there is also `Enums.LoadStatusFailedToLoad` which you should handle as a terminal state

ArcGISRuntimeSDKQt_QMLSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.qml
```diff
@@ -30,6 +30,7 @@ Rectangle {
     property OgcFeatureService featureService: null
     property FeatureLayer featureLayer : null
     property string errorMessage: ""
+    property bool loading: false
```
lsmallwood, line 33: I wonder if this could just be `property bool loading: featureService ? featureService.loadStatus === Enums.LoadStatusLoading : false`

lsmallwood (approved):
LGTM @AndrewBladon - a couple of thoughts for qml

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.cpp
```diff
@@ -98,6 +98,11 @@ void BrowseOGCAPIFeatureService::handleError(const Esri::ArcGISRuntime::Error& e
 
 void BrowseOGCAPIFeatureService::loadFeatureService(const QUrl& url)
 {
+  if (m_featureService && m_featureService->loadStatus() == LoadStatus::Loading)
+  {
+    return;
+  }
```
ldanzinger, line 104: our coding standard doesn't require brackets for one line bodys

```suggestion
  if (m_featureService && m_featureService->loadStatus() == LoadStatus::Loading)
    return;
```

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.cpp
```diff
@@ -206,10 +216,23 @@ void BrowseOGCAPIFeatureService::clearExistingFeatureLayer()
 
 void BrowseOGCAPIFeatureService::addFeatureLayerToMap()
 {
+  setLoading(false);
```
ldanzinger, line 219: It seems like there should be a simpler way where you don't need to be managing state throughout the code and setting/unsetting the loading state. Under which conditions would 'isLoading' be true? Is it just `if (m_featureService && m_featureService->loadStatus() == LoadStatus::Loading)`? Ideally we can let the API tell us when it is loading or not, but not sure if you have additional app level loading required here...

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.h
```diff
@@ -84,6 +88,7 @@ class BrowseOGCAPIFeatureService : public QObject
   QStringList m_featureCollectionList;
   Esri::ArcGISRuntime::OgcFeatureCollectionTable* m_featureCollectionTable = nullptr;
   Esri::ArcGISRuntime::FeatureLayer* m_featureLayer = nullptr;
+  bool isLoading = false;
```
ldanzinger, line 91: members should start with `m_` prefix

ArcGISRuntimeSDKQt_QMLSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.qml
AndrewBladon: There is a function called `checkForLoadingErrors` which handles the `Enums.LoadStatusFailedToLoad`:
```
function checkForLoadingErrors() {
    if (root.featureService.loadStatus === Enums.LoadStatusFailedToLoad) {
        handleError(root.featureService.loadError);
    }
}
```
Will this be sufficient?

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.qml
```diff
@@ -65,6 +65,7 @@ Item {
                 Button {
                     id: connectButton
                     text: "Load service"
+                    enabled: model.loading ? false : true;
```
ldanzinger, line 68: if you are binding to an expression that returns bool, then it is redundant to use the ternary operator:

```suggestion
                    enabled: !model.loading
```

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.qml
```diff
@@ -78,6 +79,7 @@ Item {
                 Button {
                     id: loadLayerButton
                     text: "Load selected layer"
+                    enabled: model.loading ? false : true;
```
ldanzinger, line 82:
```suggestion
                    enabled: !model.loading
```

ldanzinger (approved):
I think there should be a cleaner way to do this than to manage the bool state throughout the app. Luke had a similar suggestion as to what I was thinking. I'm ok with this going in to fix the crash, but it should be reworked to streamline the workflow afterwards

JamesMBallard:
>could you confirm that, once approved, it is ok to merge these changes into the 100.12.0 release branch?

Yes, we should fix a crash for the release. Good to merge once approved.

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.cpp
JamesMBallard: I agree with @ldanzinger. `setLoading` seems unnecessary when the API itself can tell you this information.

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseOGCAPIFeatureService/BrowseOGCAPIFeatureService.h
```diff
@@ -43,6 +43,7 @@ class BrowseOGCAPIFeatureService : public QObject
   Q_PROPERTY(QString errorMessage READ errorMessage WRITE setErrorMessage NOTIFY errorMessageChanged)
   Q_PROPERTY(QUrl featureServiceUrl READ featureServiceUrl NOTIFY urlChanged)
   Q_PROPERTY(QStringList featureCollectionList READ featureCollectionList NOTIFY featureCollectionListChanged)
+  Q_PROPERTY(bool loading READ loading NOTIFY loadingChanged)
```
JamesMBallard, line 46: This can all be inferred from `m_featureLayer` which already has a `loadStatusChanged` signal that notifies. You can simply reuse that logic as needed.

Is you are going to keep this, I would be more specific. Presumably there are a few Loadable objects in play and "loading" is a well-known API term. I would make this `featureLayerLoading` or something like that if you keep it.

JamesMBallard (commented):
There should be a simpler way to solve this that stashing bools that mimic the loadStatus of an API object. I would recommend simplifying this if it's possible in the time remaining for 100.12.

AndrewBladon:
Thanks for your suggestions @lsmallwood, @ldanzinger, and @JamesMBallard!

AndrewBladon:
@lsmallwood and @ldanzinger, thanks again for your review comments. We have made the recommended changes. Would it be possible to have the changes reviewed and approved? Thank you.

ldanzinger (approved):
looks much cleaner. Good to merge

tanneryould (approved):

lsmallwood (approved):
LGTM

PR #1275, opened 2021-08-10
Tan11389/ab/ogc feature service fix

tanneryould:
Binds the `LoadStatus` property of `m_featureService` and `m_featureLayer` to enable and disable the UI buttons rather than creating another member variable.


AndrewBladon (approved):
These changes look great @tanneryould, thanks!

PR #1282, opened 2021-08-16
C++ curve code cleanup

tanneryould:
This just removes some unnecessary header includes as well as declares the heart parts in the clockwise order they will be added rather than just by class name. I feel that helps the user understand how each part works a bit better.


AndrewBladon (approved):
Looks good @tanneryould! Thanks.

PR #1291, opened 2021-08-23
Andrewbladon/add mesh layer service issue handling

AndrewBladon:
This PR addresses issues reported with the Add Integrated Mesh Layer sample. The previous implementation of the sample resulted in unexpected behaviour if the Girona Integrated Mesh service was unavailable. These updates check to see if an error occurs while accessing the service and presents the user with an error message if applicable.

@tanneryould, @lsmallwood, and @ldanzinger - would you be able to review these changes? I would appreciate your thoughts, especially on whether you think the implemented changes are sufficient safeguarding against the service being unavailable.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Scenes/AddIntegratedMeshLayer/AddIntegratedMeshLayer.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Scenes/AddIntegratedMeshLayer/AddIntegratedMeshLayer.cpp
@@ -28,21 +28,57 @@
 using namespace Esri::ArcGISRuntime;
 
 AddIntegratedMeshLayer::AddIntegratedMeshLayer(QObject* parent /* = nullptr */):
-  QObject(parent),
-  m_scene(new Scene(this))
+  QObject(parent)
 {
   // create the integrated mesh layer
   const QUrl meshLyrUrl("https://tiles.arcgis.com/tiles/z2tnIkrLQ2BRzr6P/arcgis/rest/services/Girona_Spain/SceneServer");
-  IntegratedMeshLayer* integratedMeshLyr = new IntegratedMeshLayer(meshLyrUrl, this);
+  m_integratedMeshLyr = new IntegratedMeshLayer(meshLyrUrl, this);
 
-  // add the layer to the scene
-  m_scene->operationalLayers()->append(integratedMeshLyr);
+  // Connect the done loading signal to the handleLoadingCompleted() method.
+  connect(m_integratedMeshLyr, &IntegratedMeshLayer::doneLoading, this, &AddIntegratedMeshLayer::handleLoadingCompleted);
 
-  // set the intial viewpoint of the scene
+  // Load mesh layer
+  m_integratedMeshLyr->load();
+}
+
+void AddIntegratedMeshLayer::handleLoadingCompleted(Error error)
+{
+  if (error.isEmpty())
+  {
+    createSceneForIntegratedMesh();
+
+    m_scene->operationalLayers()->append(m_integratedMeshLyr);
+
+    setIntegratedMeshViewpoint();
+  }
+  else
+  {
+    createDefaultScene();
+
+    seterrorWhileLoading(true);
+  }
+}
+
+void AddIntegratedMeshLayer::createSceneForIntegratedMesh()
+{
+  m_scene = new Scene(this);
+  m_sceneView->setArcGISScene(m_scene);
+  emit sceneViewChanged();
+}
+
+void AddIntegratedMeshLayer::setIntegratedMeshViewpoint()
+{
   const Point initialPt(2.8259, 41.9906, 200.0, SpatialReference::wgs84());
   const Camera initialCamera(initialPt, 200.0, 190.0, 65.0, 0.0);
   const Viewpoint initialViewpoint(initialPt, initialPt.z(), initialCamera);
   m_scene->setInitialViewpoint(initialViewpoint);
+}
+
+void AddIntegratedMeshLayer::createDefaultScene()
+{
+  m_scene = new Scene(BasemapStyle::ArcGISImagery, this);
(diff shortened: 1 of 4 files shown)


ArcGISRuntimeSDKQt_CppSamples/Scenes/AddIntegratedMeshLayer/AddIntegratedMeshLayer.cpp
```diff
@@ -26,19 +26,38 @@
 #include "IntegratedMeshLayer.h"
 
 using namespace Esri::ArcGISRuntime;
+namespace
+{
+  const QUrl meshLyrUrl("https://tiles.arcgis.com/tiles/z2tnIkrLQ2BRzr6P/arcgis/rest/services/Girona_Spain/SceneServer");
```
ldanzinger, line 31: this is ok, but not sure it is necessary to have the global variable since it is only use once in the c'tor

lsmallwood (commented):
LGTM @AndrewBladon - a few comments to look at

ArcGISRuntimeSDKQt_CppSamples/Scenes/AddIntegratedMeshLayer/AddIntegratedMeshLayer.cpp
```diff
@@ -26,19 +26,38 @@
...
+
+void AddIntegratedMeshLayer::handleError(Error error)
+{
+  // If the doneLoading error message is not empty, an error has been thrown.
+  if (!error.isEmpty())
+  {
+    if (error.additionalMessage().isEmpty())
+      setErrorMessage(error.message());
+    else
+      setErrorMessage(error.message() + "\n" + error.additionalMessage());
```
ldanzinger, line 55: since this is the slot to handle error messages, it seems ideal to reasonably consolidate the logic into the slot and get rid of your setErrorMessage function.

Something like:

```suggestion

    if (error.additionalMessage().isEmpty())
      m_errorMessage = error.message();
    else
      m_errorMessage = error.message() + "\n" + error.additionalMessage();
    
    emit errorMessageChanged();
```

ArcGISRuntimeSDKQt_CppSamples/Scenes/AddIntegratedMeshLayer/AddIntegratedMeshLayer.h
```diff
@@ -33,6 +35,7 @@ class AddIntegratedMeshLayer : public QObject
   Q_OBJECT
 
   Q_PROPERTY(Esri::ArcGISRuntime::SceneQuickView* sceneView READ sceneView WRITE setSceneView NOTIFY sceneViewChanged)
+  Q_PROPERTY(QString errorMessage READ errorMessage WRITE setErrorMessage NOTIFY errorMessageChanged)
```
ldanzinger, line 38: could be simplified by using MEMBER keyword to eliminate the getter/setter

```suggestion
  Q_PROPERTY(QString errorMessage MEMBER m_errorMessage NOTIFY errorMessageChanged)
```

ArcGISRuntimeSDKQt_CppSamples/Scenes/AddIntegratedMeshLayer/AddIntegratedMeshLayer.h
```diff
@@ -42,13 +45,21 @@ class AddIntegratedMeshLayer : public QObject
 
 signals:
   void sceneViewChanged();
+  void errorMessageChanged();
 
 private:
   Esri::ArcGISRuntime::SceneQuickView* sceneView() const;
   void setSceneView(Esri::ArcGISRuntime::SceneQuickView* sceneView);
+  QString errorMessage() const;
+  void setErrorMessage(QString message);
```
ldanzinger, line 54: if you decide to keep this, it should be const &

ArcGISRuntimeSDKQt_QMLSamples/Scenes/AddIntegratedMeshLayer/AddIntegratedMeshLayer.qml
```diff
@@ -16,13 +16,16 @@
 
 import QtQuick 2.6
 import Esri.ArcGISRuntime 100.12
+import QtQuick.Dialogs 1.2
```
ldanzinger, line 19: same comment here - I suggest using the QtQuickControls 2 dialog instead and removing this dependency

ldanzinger (changes requested):

tanneryould (approved):
LGTM

ArcGISRuntimeSDKQt_CppSamples/Scenes/AddIntegratedMeshLayer/AddIntegratedMeshLayer.qml
```diff
@@ -17,6 +17,7 @@
 import QtQuick 2.6
 import QtQuick.Controls 2.2
 import Esri.Samples 1.0
+import QtQuick.Controls 2.5
```
ldanzinger, line 20: controls 2.2 is imported above. should be able to just update that instead of importing 2 different versions

ldanzinger (approved):

ArcGISRuntimeSDKQt_QMLSamples/Scenes/AddIntegratedMeshLayer/AddIntegratedMeshLayer.qml
```diff
@@ -15,7 +15,7 @@
 // [Legal]
 
 import QtQuick 2.6
-import Esri.ArcGISRuntime 100.12
+import Esri.ArcGISRuntime 100.13
```
tanneryould, line 18:
```suggestion
import Esri.ArcGISRuntime 100.12
```

PR #1298, opened 2021-09-08
Revisit OpenStreetMapLayer sample 

Gela:
This PR updates the existing OpenStreetMapLayer sample. It previously used an OSM basemap, but the original intention of this sample was to actually demonstrate use of the OpenStreetMapLayer class.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Layers/OSM_Layer/OSM_Layer.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Layers/OSM_Layer/OSM_Layer.cpp
@@ -46,8 +46,14 @@ void OSM_Layer::componentComplete()
   // find QML MapView component
   m_mapView = findChild<MapQuickView*>("mapView");
 
-  // Create a map using the OpenStreetMap basemap
-  m_map = new Map(BasemapStyle::OsmStandard, this);
+  // Create a map
+  m_map = new Map();
+
+
+  // Create a new OpenStreetMapLayer and add it to the list of operational layers
+  OpenStreetMapLayer* osm = new OpenStreetMapLayer();
+  m_map->operationalLayers()->append(osm);
+
 
   // Set map to map view
   m_mapView->setMap(m_map);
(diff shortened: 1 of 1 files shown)


AndrewBladon (approved):
Looks good @Gela!

ArcGISRuntimeSDKQt_CppSamples/Layers/OSM_Layer/OSM_Layer.cpp
```diff
@@ -46,8 +46,13 @@ void OSM_Layer::componentComplete()
...
   m_mapView = findChild<MapQuickView*>("mapView");
 
-  // Create a map using the OpenStreetMap basemap
-  m_map = new Map(BasemapStyle::OsmStandard, this);
+  // Create a map
+  m_map = new Map();
+
+  // Create a new OpenStreetMapLayer and add it to the list of operational layers
+  OpenStreetMapLayer* osm = new OpenStreetMapLayer();
+  m_map->operationalLayers()->append(osm);
```
tanneryould, line 54: Because OpenStreetMapLayer does not support identify, query, selection, or time, I would suggest we add it to the base layers instead. This way it'll be a little more clear as to the intended purpose of this class.

We should also set `m_map`'s and `osm`'s parent to `this` so it will be available through the runtime of the sample. Generally we set a class's parent to `this` any time `new` is used.
```suggestion
  m_map = new Map(this);

  // Create a new OpenStreetMapLayer and add it to the list of basemap base layers
  OpenStreetMapLayer* osm = new OpenStreetMapLayer(this);
  m_map->basemap()->baseLayers()->append(osm);
```

tanneryould (changes requested):
Looks good, just two small things.

tanneryould:
Oh @Gela I believe this PR should also target `v.next` and not `main`

Gela:
Closing this PR since it's against the fork of the original repo and against the incorrect branch. New PR can be viewed here: https://github.com/Esri/arcgis-runtime-samples-qt/pull/1301

PR #1300, opened 2021-09-08
Update Token Authentication Sample

AndrewBladon:
This PR changes the authentication method of the Token Authentication sample. Previously, the sample accessed a protected data service using a URL and added this to the `Map` as a `MapImageLayer`. The updated sample now accesses ArcGIS online using a `Portal`, creates a `PortalItem` using an Item ID, and instantiates the map using the `PortalItem`. The changes to the implementation mean that the protected resource must now be accessed using an AGOL login, rather than username = "user1" and password = "user1".

In both the previous and updated implementations the samples use the `AuthenticationView` QML item from the toolkit. When an attempt is made to access a protected resource, the Authentication view handles the authentication process automatically.

The README files have also been updated to be consistent with other samples and the changes in this PR.


ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/TokenAuthentication/TokenAuthentication.cpp
```diff
@@ -50,22 +49,15 @@ void TokenAuthentication::componentComplete()
...
   if (!m_mapView)
-      return;
+    return;
 
   m_mapView->setWrapAroundMode(WrapAroundMode::Disabled);
 
-  // Create a map using the topographic basemap
-  m_map = new Map(BasemapStyle::ArcGISTopographic, this);
-  m_map->setInitialViewpoint(Viewpoint(Envelope(-19221397.32591107, -208306.59714691807, -3024411.4154897667, 11939432.83566906, SpatialReference(3857))));
+  m_portal = new Portal(m_portalURL, this);
```
lsmallwood, line 56: nit-pick. there's no reason to wait for component complete to setup `m_portal`, `m_portalItem` and `m_map`. These could all be done in the initializer list. That way they'll be created ready to bet set to the map view in this method.

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/TokenAuthentication/TokenAuthentication.h
```diff
@@ -38,11 +40,14 @@ class TokenAuthentication : public QQuickItem
 
   void componentComplete() override;
   static void init();
-  Q_INVOKABLE void loadSecuredLayer();
 
 private:
+  QUrl m_portalURL = QUrl("https://www.arcgis.com/");
```
lsmallwood, line 45: optional. these could be moved to the cpp file in an anonymous namespace rather than having them as state for the class.

lsmallwood (approved):
Looks good to me @AndrewBladon - nice clean-up. A couple of optional things to look at but no need to re-review.

Gela (approved):
No additional feedback, looks good

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/TokenAuthentication/TokenAuthentication.h
tanneryould: Agreed, this can just be set in the .cpp file. Furthermore, I'm not actually certain it needs to be a variable because it's only referenced once. Also arcgis.com is the default portal url so it really doesn't need to be passed at all.

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/TokenAuthentication/TokenAuthentication.h
```diff
@@ -38,11 +40,14 @@ class TokenAuthentication : public QQuickItem
...
+  QString m_itemID = "e5039444ef3c48b8a8fdc9227f9be7c1";
   Esri::ArcGISRuntime::Map* m_map = nullptr;
   Esri::ArcGISRuntime::MapQuickView* m_mapView = nullptr;
+  Esri::ArcGISRuntime::Portal* m_portal = nullptr;
+  Esri::ArcGISRuntime::PortalItem* m_portalItem = nullptr;
```
tanneryould, line 50: These shouldn't need to be included in the header file. Take a look at how I did it with the Geotriggers sample: https://github.com/Esri/arcgis-runtime-samples-qt/blob/v.next/ArcGISRuntimeSDKQt_CppSamples/Analysis/Geotriggers/Geotriggers.cpp#L51

tanneryould (approved):
LGTM - though a couple of small notes others have mentioned

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/TokenAuthentication/TokenAuthentication.h
JamesMBallard: @tanneryould is your concern having these as class members, or initializing them to `nullptr`?

tanneryould: @JamesMBallard having these as class members - is there a point to declaring all of these members instead of just writing the following?
```cpp
m_map = new Map(new Portal("e5039444ef3c48b8a8fdc9227f9be7c1", this), this);
```

JamesMBallard: @tanneryould good point. That makes sense to me to just utilize them where needed, once. Thanks for clarifying.

lsmallwood: I think avoiding having these as members is a good suggestion but I would still recommend naming the variables like the portal item Id and also the url for arcgis.com. We want to make sure the code is easy to read and straightforward for users to swap in there own portals/items.

AndrewBladon: Thanks for your comments and suggestions @tanneryould, @JamesMBallard, and @lsmallwood. I have kept the portalURL, itemID, m_portal, and m_portalItem variables. Although this is not the most efficient implementation, I think explicitly declaring each of the these as variables helps demonstrate how the authentication is set-up and makes the code readable and re-usable.

AndrewBladon: @tanneryould and @JamesMBallard, following my comment above and merging this PR, I discussed your suggestions with @lsmallwood and realised I had not fully understood them. To remedy this, I have made some further changes to remove m_portal and m_portalItem from the member variables. These changes can be found in https://github.com/Esri/arcgis-runtime-samples-qt/pull/1303

The `Portal` and `PortalItem` objects are no longer member variables, and are created as local variables in the constructor. These local variables are then used to create the `Map` object. Hopefully the changes in the above PR are consistent with your suggestions.

Thanks again!

kathleen1:
Build 3289, C++ & QML
- [x] Windows
- [ ] macOS
- [ ] iOS
- [ ] Linux
- [ ] Android

PR #1302, opened 2021-09-13
Tan11389/enable sample viewer keyboard nav

tanneryould:
Adds code to force focus onto the map/scene views so that keyboard navigation works in the sample viewer and after clicking on a different QML component. This also enables the delete key in the "Sketch on map" samples within the sample viewers. Did some spot checking on Mac and everything looks ok.


AndrewBladon (approved):
Looks good @tanneryould! 

I have a small suggestion - it might be useful to have a comment before the new code block that provides an explanation of its purpose. With this understanding and context, a user can decide whether they need/want to include something similar in their apps.

Gela (approved):

ldanzinger:
Couple general questions:

* `onFocusChanged: focus = true;` seems potentially problematic as it could be circular or cause focus to fight with other things that are trying to get or need focus. Are we sure it doesn't cause UI fighting between components in all cases for the changed files? For example, the sample that synchronizes the 2D map and 3D scene viewpoints - if both are forcing focus, 
* I think it definitely makes sense to add this for the sketch sample (or anywhere else where we know we need it), but is it needed for all samples? Seems like there could potentially be fallout here, but I haven't tested to confirm that

PR #1304, opened 2021-09-16
Remove duplicate Component.onCompleted properties

tanneryould:
Some views had duplicate `Component.onCompleted` properties after the template update. This fixes that issue.


MicahSamaduroff (approved):
Looks good to me

ldanzinger (approved):

PR #1312, opened 2021-09-30
Update relevant API checks

tanneryould:
This updates the Relevant API section of the README checker to be a little more useful.


Gela (approved):
Thanks for fixing this

PR #1328, opened 2021-10-29
Update WMTS_Layer.qml

ldanzinger:
Jason found that this sample is not working. Looks like the function name got mangled


tanneryould (approved):

tanneryould:
Good catch, thank you

PR #1331, opened 2021-12-01
Reorder metadata snippets

tanneryould:
This PR updates the order of the snippets in C++ metadata files, so that the primary C++ file is first, and the primary QML file is last. This way, the website will display C++ files upon load rather than QML files, for our C++ samples.


AndrewBladon (approved):
@tanneryould, I browsed through the changed files and randomly selected 20-30 of them to review. They all looked good to me!

ldanzinger (approved):
LGTM

Gela (approved):

PR #1336, opened 2021-12-15
Update Runtime version from 100.13 to 100.14

AndrewBladon:
This PR updates the Runtime version from 100.13 to 100.14 in the Display overview map sample files.

The files for the sample were created using Runtime version 100.13 and the new sample files were pushed to the repository immediately before the Runtime version was updated in all sample files. As such, the Runtime version in this sample needed updating from 100.13 to 100.14 to be consistent with the rest of the repository.


mfeigl (approved):
LGTM, thanks for the changes

PR #1340, opened 2021-12-31
Fix these broken properties.

JamesMBallard:
Hi all. I don't know who is around to do reviews at the moment, but we need to fix this sample asap. This is causing problems with the iOS samples build.

You cannot have write-only `Q_PROPERTY`, those are illegal and result in these, which breaks the iOS build
```
Warning: Property declaration dimensionLayerVisibility has no READ accessor function or associated MEMBER variable. The property will be invalid
```

Summary of changes:
 * Added property getters and missing `READ`.
 * Some of the naming is suboptimal, but this is subjective. I changed these property names
   * `dimensionLayerVisibility` to `dimensionLayerVisible`
   * `definitionExpressionApplied`  to `useDefinitionExpression`
 * Change `const bool` args to `bool`
 * Rename `setdimensionsAvailable` to `setDimensionsAvailable`

This fixes the iOS build. I also tested the changes locally and the sample works.


jared-2016 (approved):
Looks good! 👍

PR #1343, opened 2022-01-31
Guil8553/7030 clazy fixes

GuillaumeBelz:
Here some fixes for clang-tidy and clazy warnings.

Warnings:
- const with Qt's containers
- remove `QScopedPointer`: we can create an issue for that
- `this` in connection parameters
- using `UniqueConnection` with non-member

Some error are not obvious, thrust the tools!


ArcGISRuntimeSDKQt_CppSamples/Analysis/AnalyzeViewshed/AnalyzeViewshed.cpp
```diff
@@ -213,7 +213,8 @@ void AnalyzeViewshed::calculateViewshed()
 void AnalyzeViewshed::processResults(GeoprocessingResult *results)
 {
   // Get the results from the outputs as GeoprocessingFeatures
-  GeoprocessingFeatures* viewshedResultFeatures = static_cast<GeoprocessingFeatures*>(results->outputs()["Viewshed_Result"]);
```
GuillaumeBelz, line 216: This is another example of using a non-const Qt's container with a non-const function. In this code, `results->outputs()` is not const then `operator[]` is not const and the container is detached (deep copy).

To solve that, it's necessary to force `const`:
- using `qAsConst`
- or creating a local const variable

The difficulty with `qAsConst` in this code is it must be used with the objet, not the container:
```cpp
... static_cast<GeoprocessingFeatures*>(qAsConst(results)->outputs()["Viewshed_Result"]);
// not
// ... static_cast<GeoprocessingFeatures*>(qAsConst(results->outputs())["Viewshed_Result"]);
```

I don't know if it's clear that `const` and `qAsConst` are used here to avoid to detach the container. Which syntaxe do you prefer?

ArcGISRuntimeSDKQt_CppSamples/Analysis/Geotriggers/Geotriggers.cpp
```diff
@@ -151,7 +151,7 @@ void Geotriggers::createGeotriggerMonitor(ServiceFeatureTable* serviceFeatureTab
 void Geotriggers::handleGeotriggerNotification(GeotriggerNotificationInfo* geotriggerNotificationInfo)
 {
   // Create a QScopedPointer so GeotriggerNotificationInfo does not persist outside this scope
-  QScopedPointer<GeotriggerNotificationInfo> scopedNotification(geotriggerNotificationInfo);
+  std::unique_ptr<GeotriggerNotificationInfo> scopedNotification(geotriggerNotificationInfo);
```
GuillaumeBelz, line 154: Probably no reason to continue to use `QScopedPointer`.

ArcGISRuntimeSDKQt_CppSamples/Analysis/Geotriggers/Geotriggers.cpp
```diff
@@ -183,7 +183,7 @@ void Geotriggers::handleGeotriggerNotification(GeotriggerNotificationInfo* geotr
     m_currentFeaturesEntered[geotriggerName].removeAll(featureName);
   }
 
-  displayInfoChanged();
+  emit displayInfoChanged();
```
GuillaumeBelz, line 186: Missing keyword `emit` for signals. (I've missed something about the use of `emit`?)

ArcGISRuntimeSDKQt_CppSamples/Analysis/StatisticalQuery/StatisticalQuery.cpp
```diff
@@ -69,7 +69,7 @@ void StatisticalQuery::componentComplete()
   // Set map to map view
   m_mapView->setMap(m_map);
 
-  connect(m_featureTable, &ServiceFeatureTable::errorOccurred, [this](Error e)
+  connect(m_featureTable, &ServiceFeatureTable::errorOccurred, this, [this](Error e)
```
GuillaumeBelz, line 72: I suppose that `m_featureTable` is deleted in the same time as `this` and then there is no risk that the connection is still alive with invalid `this` pointer. But it's easy to do a mistake.

And it's better to be consistent in the code and always have a third parameter.

ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/SearchForWebmap/SearchForWebmap.cpp
```diff
@@ -96,7 +96,7 @@ QString SearchForWebmap::mapLoadError() const
   return m_mapLoadeError;
 }
 
-void SearchForWebmap::search(const QString keyword)
+void SearchForWebmap::search(const QString& keyword)
```
GuillaumeBelz, line 99: Missing reference in parameter?

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ControlAnnotationSublayerVisibility/ControlAnnotationSublayerVisibility.cpp
```diff
@@ -55,7 +55,7 @@ QString defaultDataPath()
 } // namespace
 
 // sample MMPK location
-const QString sampleFileAnno {"/ArcGIS/Runtime/Data/mmpk/GasDeviceAnno.mmpk"};
+const QString sampleFileAnno {"/ArcGIS/Runtime/Data/mmpk/GasDeviceAnno.mmpk"}; // why not in namespace?
```
GuillaumeBelz, line 58: I don't know if there is a reason why this variable is outside anonyme namespace.

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/GODictionaryRenderer/GODictionaryRenderer.cpp
```diff
@@ -94,6 +94,8 @@ void GODictionaryRenderer::componentComplete()
   m_graphicsOverlay->setRenderer(renderer);
   //! [Apply Dictionary Renderer Graphics Overlay Cpp]
 
+  Q_UNUSED(specType);
```
GuillaumeBelz, line 97: Not important, but there is a unused variable in snippet. This is just to remove the warning.

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/GOSymbols/GOSymbols.cpp
```diff
@@ -103,7 +103,7 @@ void GOSymbols::addBuoyPoints(GraphicsOverlay* graphicsOverlay)
 
   // create the symbology for the points
   constexpr float size = 10.0f;
-  SimpleMarkerSymbol* sms = new SimpleMarkerSymbol(SimpleMarkerSymbolStyle::Circle, QColor("red"), size, this);
+  SimpleMarkerSymbol* sms = new SimpleMarkerSymbol(SimpleMarkerSymbolStyle::Circle, QColor("red"), size, this); // leak?
```
GuillaumeBelz, line 106: That is a crazy warning from clazy. I guess clazy suppose that the range-for loop can have 0 loop, which is not true as the containers is const. I think this is a bug in clazy.

ArcGISRuntimeSDKQt_CppSamples/Geometry/DensifyAndGeneralize/DensifyAndGeneralize.cpp
```diff
@@ -126,11 +126,12 @@ void DensifyAndGeneralize::updateGeometry(bool densify, double maxSegmentLength,
   m_resultLineGraphic->setGeometry(polyline);
 
   // Update the multipoint graphic
-  MultipointBuilder multipointBuilder(polyline.spatialReference());
-  PointCollection* pointCollection = new PointCollection(polyline.spatialReference(), this);
```
GuillaumeBelz, line 130: This is not a leak, because `this` is the parent, but if the parts is empty, this function returns without deleting the point collection.

ArcGISRuntimeSDKQt_CppSamples/Geometry/FormatCoordinates/FormatCoordinates.cpp
```diff
@@ -180,7 +180,7 @@ void FormatCoordinates::setMapView(MapQuickView* mapView)
     // using the point, refresh the graphic and the text
     handleLocationUpdate(std::move(mapPoint));
   },
-  Qt::UniqueConnection);
+  Qt::UniqueConnection); // Finds usages of Qt::UniqueConnection when the slot is a functor, lambda or non-member function. That connect() overload does not support Qt::UniqueConnection.
```
GuillaumeBelz, line 183: I didn't know about that. I guess we should create an issue to work on that.

ArcGISRuntimeSDKQt_CppSamples/Layers/BrowseWfsLayers/BrowseWfsLayers.cpp
```diff
@@ -169,7 +168,9 @@ void BrowseWfsLayers::addFeatureLayerToMap()
     qDebug() << "Error! No Renderer defined for this GeometryType";
     return;
   }
+
   // apply renderer to layer
+  FeatureLayer* featureLayer = new FeatureLayer(m_wfsFeatureTable, this);
```
GuillaumeBelz, line 173: Reorder the creation of `FeatureLayer` to avoid leak.

ArcGISRuntimeSDKQt_CppSamples/Layers/DisplayDimensions/DisplayDimensions.h
```diff
@@ -38,8 +38,8 @@ class DisplayDimensions : public QObject
   Q_PROPERTY(Esri::ArcGISRuntime::MapQuickView* mapView READ mapView WRITE setMapView NOTIFY mapViewChanged)
   Q_PROPERTY(QString errorMessage READ errorMessage WRITE setErrorMessage NOTIFY errorMessageChanged)
   Q_PROPERTY(QString dimensionLayerName READ dimensionLayerName WRITE setDimensionLayerName NOTIFY dimensionLayerNameChanged)
-  Q_PROPERTY(bool dimensionLayerVisible READ dimensionLayerVisible WRITE setDimensionLayerVisible)
-  Q_PROPERTY(bool useDefinitionExpression READ useDefinitionExpression WRITE setUseDefinitionExpression)
+  Q_PROPERTY(bool dimensionLayerVisible READ dimensionLayerVisible WRITE setDimensionLayerVisible) // no NOTIFY?
```
GuillaumeBelz, line 41: Properties should have `CONSTANT` or `NOTIFY`

ArcGISRuntimeSDKQt_CppSamples/Maps/ApplyScheduledMapUpdates/ApplyScheduledMapUpdates.cpp
```diff
@@ -221,14 +222,14 @@ void ApplyScheduledMapUpdates::connectSyncSignals()
...
-    m_syncJobConnection = connect(m_syncJob, &OfflineMapSyncJob::jobDone, [this]()
+    m_syncJobConnection = connect(m_syncJob, &OfflineMapSyncJob::jobDone, this, [this]()
     {
       if (m_syncJob->result()->isMobileMapPackageReopenRequired())
       {
         // close mmpk out
         m_mobileMapPackage->close();
-        delete m_mobileMapPackage;
         disconnect(m_mobileMapPackage, &MobileMapPackage::doneLoading, this, &ApplyScheduledMapUpdates::onMmpkDoneLoading);
+        delete m_mobileMapPackage;
```
GuillaumeBelz, line 232: Is it necessary to disconnect manually?

ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap_Overrides/GenerateOfflineMap_Overrides.cpp
```diff
@@ -130,7 +130,7 @@ void GenerateOfflineMap_Overrides::setBasemapLOD(int min, int max)
     return;
 
   LayerListModel* layers = m_map->basemap()->baseLayers();
-  if (layers && layers->size() < 1)
+  if (!layers || layers->isEmpty())
```
GuillaumeBelz, line 133: Modify the condition to return if `layer` is null.

ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap_Overrides/GenerateOfflineMap_Overrides.cpp
```diff
@@ -357,6 +357,8 @@ void GenerateOfflineMap_Overrides::takeMapOffline()
       emit hideWindow(1500, true);
       m_mapView->setMap(generateJob->result()->offlineMap(this));
       break;
+    default: // do nothing
```
GuillaumeBelz, line 360: To remove warning.

ArcGISRuntimeSDKQt_CppSamples/Maps/ManageBookmarks/ManageBookmarks.cpp
```diff
@@ -95,14 +95,15 @@ void ManageBookmarks::createInitialBookmarks()
...
-  Bookmark* bookmark = new Bookmark(name, viewpoint, this);
-
-  // Add it to the map's bookmark list
+  // Get the bookmark list
   BookmarkListModel* bookmarks = dynamic_cast<BookmarkListModel*>(m_bookmarks);
   if (!bookmarks)
     return;
 
+  // Create the bookmark from the name and viewpoint
+  Bookmark* bookmark = new Bookmark(name, viewpoint, this);
```
GuillaumeBelz, line 104: Call `new` after the `return`

ArcGISRuntimeSDKQt_CppSamples/Scenes/ExtrudeGraphics/ExtrudeGraphics.cpp
```diff
@@ -97,7 +97,7 @@ void ExtrudeGraphics::componentComplete()
   graphicsOverlay->setRenderer(renderer);
 
   // setup graphic locations
-  QList<Point> pointsList;
+  QVector<Point> pointsList;
```
GuillaumeBelz, line 100: In Qt6, this is not important, but in Qt5, we should use vector as default. (Use `std::vector` here?)

ArcGISRuntimeSDKQt_CppSamples/Scenes/ExtrudeGraphics/ExtrudeGraphics.cpp
```diff
@@ -107,8 +107,9 @@ void ExtrudeGraphics::componentComplete()
   for (const Point& point : pointsList)
   {
     // create a random z value
-    int randNum = rand() % 6 + 1;
-    double z = m_maxZ * randNum;
+    const int randNum = rand() % 6 + 1; // security.insecureAPI.rand: Warn on uses of inferior random number generating functions (only if arc4random function is available): drand48, erand48, jrand48, lcong48, lrand48, mrand48, nrand48, random, rand_r.
```
GuillaumeBelz, line 110: // security.insecureAPI.rand: Warn on uses of inferior random number generating functions (only if arc4random function is available): drand48, erand48, jrand48, lcong48, lrand48, mrand48, nrand48, random, rand_r.

ArcGISRuntimeSDKQt_CppSamples/Search/FindPlace/FindPlace.h
```diff
@@ -59,7 +59,7 @@ class FindPlace : public QQuickItem
   Q_INVOKABLE void setSuggestionsText(const QString& searchText);
   Q_INVOKABLE void geocodePOIs(const QString& poi, const QString& location);
   Q_INVOKABLE void geocodePOIs(const QString& poi, const Esri::ArcGISRuntime::Point& location);
-  Q_INVOKABLE void geocodePOIs(const QString& poi, SearchMode mode);
+  Q_INVOKABLE void geocodePOIs(const QString& poi, FindPlace::SearchMode mode);
```
GuillaumeBelz, line 62: In signals, slots and invokable functions, the types must be full qualified.

ArcGISRuntimeSDKQt_CppSamples/UtilityNetwork/CreateLoadReport/CreateLoadReport.cpp
```diff
@@ -233,21 +233,19 @@ void CreateLoadReport::runReport(const QStringList& selectedPhaseNames)
...
   UtilityNetworkAttributeComparison* utilityNetworkAttributeComparison =
         new UtilityNetworkAttributeComparison(
           m_phasesCurrentAttribute,
           UtilityAttributeComparisonOperator::DoesNotIncludeAny,
           codedValue.code(), this);
 
-  UtilityTraceOrCondition* utilityTraceOrCondition = nullptr;
-
   // Chain it with the base condition using an OR operator.
-  if (utilityNetworkAttributeComparison && m_baseCondition)
```
GuillaumeBelz, line 246: If `baseCondition` is false, `utilityTraceOrCondition` is not created and the function return without using `utilityNetworkAttributeComparison`. Test `baseCondition` in the beginning of the function to avoid the "leak".

ArcGISRuntimeSDKQt_CppSamples/UtilityNetwork/CreateLoadReport/CreateLoadReport.h
```diff
@@ -54,7 +54,6 @@ class CreateLoadReport : public QObject
   static void init();
   Q_INVOKABLE void addPhase(const QString& phaseToAdd);
   Q_INVOKABLE void runReport(const QStringList& selectedPhaseNames);
-  Q_ENUMS(SampleStatus)
```
GuillaumeBelz, line 57: `Q_ENUMS` is deprecated

ArcGISRuntimeSDKQt_CppSamples/UtilityNetwork/PerformValveIsolationTrace/PerformValveIsolationTrace.cpp
```diff
@@ -86,15 +86,15 @@ struct ScopedCleanup
...
   QObject(parent),
   m_map(new Map(BasemapStyle::ArcGISStreetsNight, this)),
+  m_cred(new Credential{sampleServer7Username, sampleServer7Password, this}),
   m_startingLocationOverlay(new GraphicsOverlay(this)),
   m_filterBarriersOverlay(new GraphicsOverlay(this)),
-  m_cred(new Credential{sampleServer7Username, sampleServer7Password, this}),
-  m_graphicParent(new QObject()),
-  m_serviceGeodatabase(new ServiceGeodatabase(featureServiceUrl, m_cred, this))
+  m_serviceGeodatabase(new ServiceGeodatabase(featureServiceUrl, m_cred, this)),
+  m_graphicParent(new QObject())
```
GuillaumeBelz, line 93: Initialization in same order than declaration.

ArcGISRuntimeSDKQt_CppSamples/Analysis/AnalyzeViewshed/AnalyzeViewshed.cpp
JamesMBallard: I like your proposed solution of creating a const local.

ArcGISRuntimeSDKQt_CppSamples/Scenes/ExtrudeGraphics/ExtrudeGraphics.cpp
JamesMBallard: I have no recommendations here. Sounds like we should fix the warnings, but I will defer to the team on recommendations.

ArcGISRuntimeSDKQt_CppSamples/UtilityNetwork/CreateLoadReport/CreateLoadReport.cpp
JamesMBallard: Again, there is no leak (even with quotes, "leak") since there is a parent. But good catch that we can rewrite the logic to avoid a heap allocation altogether in this case.

ArcGISRuntimeSDKQt_CppSamples/UtilityNetwork/PerformValveIsolationTrace/PerformValveIsolationTrace.cpp
```diff
@@ -86,15 +86,15 @@ struct ScopedCleanup
...
 {
   // disable UI while loading service geodatabase and utility network
   m_tasksRunning = true;
-  emit tasksRunningChanged();
+  // emit tasksRunningChanged(); // emit inside constructor?
```
JamesMBallard, line 97: Remove it.

JamesMBallard (commented):
Most of the recommendations are good, but some require some investigation. See comments. We should get Tanner and Gela involved in the review as well.

ArcGISRuntimeSDKQt_CppSamples/Analysis/Geotriggers/Geotriggers.cpp
GuillaumeBelz: no, it's not deprecated. And until we don't want to move the pointer, it's not bad to use `QScopedPointer`. I can create an issue in sample repo.

ArcGISRuntimeSDKQt_CppSamples/Analysis/Geotriggers/Geotriggers.cpp
GuillaumeBelz: A comment about this one: `emit` is an empty `#define`, it does nothing. Then it's not an error to not use `emit`.

But I think we should use `emit`:
- to be consistent with the existing code
- to improve readability of signal emitting

I guess if we want to remove `emit` keyword, we should create an issue to talk about this guideline and remove `emit` everywhere in the code.

See https://www.kdab.com/nailing-13-signal-slot-mistakes-clazy-1-3/ about this clazy warning.

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/GOSymbols/GOSymbols.cpp
GuillaumeBelz: I think this is a mistake of clang-tidy. The pattern is:
```cpp
p = new...;
for (...) {
  // use p
}
```
In this case, clang-tidy think the loop can be empty and then `p` is not used. And then there is a leak.

In our code, the loop runs over a const non-empty container and there is no leak. But clang-tidy doesn't know the loop is never empty.

We can ignore the warning in this case.

ArcGISRuntimeSDKQt_CppSamples/Geometry/FormatCoordinates/FormatCoordinates.cpp
GuillaumeBelz: I guess the best is to remove that, as this is ignored for the moment.

ArcGISRuntimeSDKQt_CppSamples/Maps/ApplyScheduledMapUpdates/ApplyScheduledMapUpdates.cpp
GuillaumeBelz: I removed the disconnect.

ArcGISRuntimeSDKQt_CppSamples/Scenes/ExtrudeGraphics/ExtrudeGraphics.cpp
GuillaumeBelz: I keep like that. I guess we can ignore this warning.

ArcGISRuntimeSDKQt_CppSamples/Analysis/Geotriggers/Geotriggers.cpp
JamesMBallard: We should keep using `emit`, even if it technically does nothing. It adds a lot to the readability of the code, and informs people that we are emitting a signal as opposed to calling a function.

I agree with the clazy doc:
>11. incorrect-emit
>For readability purposes you should always use emit (or Q_EMIT) when calling a signal. Conversely, you should not use those macros when calling something other than a signal.

ArcGISRuntimeSDKQt_CppSamples/Geometry/FormatCoordinates/FormatCoordinates.cpp
JamesMBallard: The samples team needs to follow up on this. If the logic of the sample depends on this only ever being connected once, then it will not work correctly. @tanneryould please follow up with @GuillaumeBelz and log an issue if you haven't already.

ArcGISRuntimeSDKQt_CppSamples/Geometry/FormatCoordinates/FormatCoordinates.cpp
```diff
@@ -179,8 +179,7 @@ void FormatCoordinates::setMapView(MapQuickView* mapView)
 
     // using the point, refresh the graphic and the text
     handleLocationUpdate(std::move(mapPoint));
-  },
-  Qt::UniqueConnection); // lambda-unique-connection: Finds usages of Qt::UniqueConnection when the slot is a functor, lambda or non-member function. That connect() overload does not support Qt::UniqueConnection.
+  });
```
JamesMBallard, line 182: As indicated I'd like the samples team to follow up on this. Removing it is good since it won't show the clazy code anymore, and it's not working now anyways.

cc: @tanneryould.

JamesMBallard (approved):

tanneryould (approved):
LGTM and I created an issue in the common-samples repo to follow up on @GuillaumeBelz and @JamesMBallard's comments

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/ShowPopup/ShowPopup.cpp
```diff
@@ -86,7 +86,8 @@ void ShowPopup::onIdentifyLayerCompleted(QUuid, IdentifyLayerResult* rawIdentify
   }
 
   m_featureLayer->clearSelection();
-  for (GeoElement* geoElement : identifyResult->geoElements())
+  const auto elements = identifyResult->geoElements();
+  for (GeoElement* geoElement : elements)
```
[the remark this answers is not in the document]
JamesMBallard: Yes. Qt uses implicit sharing for many containers and QList is one of them. Clazy is a static analysis tool that flags up potential Qt-specific pitfalls, and this is one of them.

Here's the check: https://github.com/KDE/clazy/blob/master/docs/checks/README-range-loop-detach.md
https://doc.qt.io/qt-5/implicit-sharing.html#implicit-sharing-in-detail

The risk here is that the QList in the loop might detach and deep-copy the list contents. By making the list constant, we can make sure it won't detach.

ArcGISRuntimeSDKQt_CppSamples/Geometry/FormatCoordinates/FormatCoordinates.cpp
JamesMBallard: No. There is nuance to this. `Qt::UniqueConnection` does not work with lambda style connections, so the task is to:
 * Determine if a UniqueConnection is actually needed here. Is there any chance this will be connected multiple times?
 * If that won't happen, remove `Qt::UniqueConnection` since it does not do anything here anyways.
 * If this will be connected more than once, we need to update the sample code to use some other logic to make sure this connection is not made more than once.

In the short term, it is safe to remove `Qt::UniqueConnection` now since it won't work as written.

GuillaumeBelz:
https://runtime-qt.esri.com/view/vTest/job/vtest/job/source-interface/304/downstreambuildview/

PR #1344, opened 2022-02-02
add connection to thumbnailUrl

GuillaumeBelz:
Add connection to thumbnailUrl.


tanneryould (approved):
I trust your judgement, but I'm generally unclear on the difference between the sample's own `thumbnailUrl` `Q_PROPERTY` and that of `PortalUser`.

PR #1346, opened 2022-02-09
FilterByDefinitionExpressionOrDisplayFilter: wait for draw status complete before reporting back the count

Gela:
This pull request addresses an issue with this sample where if the user sets a definition expression and then hits reset, it would update the new feature count too quickly before the map has finished drawing, therefore reporting a number that is lower than expected. To address this, we connect a lambda to listen for drawing complete before reporting back anything.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Features/FilterByDefinitionExpressionOrDisplayFilter/FilterByDefinitionExpressionOrDisplayFilter.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Features/FilterByDefinitionExpressionOrDisplayFilter/FilterByDefinitionExpressionOrDisplayFilter.cpp
@@ -33,15 +33,9 @@
 #include "ServiceFeatureTable.h"
 #include "Viewpoint.h"
 
-#include <chrono>
-#include <thread>
-
 #include <QUrl>
 
 using namespace Esri::ArcGISRuntime;
-//using namespace std::this_thread;     // sleep_for, sleep_until
-//using namespace std::chrono_literals; // ns, us, ms, s, h, etc.
-//using std::chrono::system_clock;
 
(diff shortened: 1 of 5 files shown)


ArcGISRuntimeSDKQt_CppSamples/Features/FilterByDefinitionExpressionOrDisplayFilter/FilterByDefinitionExpressionOrDisplayFilter.cpp
```diff
@@ -133,3 +133,32 @@ void FilterByDefinitionExpressionOrDisplayFilter::queryFeatureCountInCurrentExte
...
+void FilterByDefinitionExpressionOrDisplayFilter::resetDefExpressionParams()
+{
+  m_featureLayer->setDefinitionExpression("");
+
+  connect(m_mapView, &MapQuickView::drawStatusChanged, this, [this](DrawStatus drawStatus)
+  {
+    drawStatus == DrawStatus::InProgress ? m_mapDrawing = true : m_mapDrawing = false;
+    emit mapDrawStatusChanged();
+
+    queryFeatureCountInCurrentExtent();
```
tanneryould, line 157: Is there a reason `queryFeatureCountInCurrentExtent()` should run when the map is still drawing?
```suggestion
    if (drawStatus == DrawStatus::Completed)
    {
      m_mapDrawing = false;
      queryFeatureCountInCurrentExtent();
    }
    else
    {
      m_mapDrawing = true;
    }
    
    emit mapDrawStatusChanged();
```

tanneryould (changes requested):
The C++ implementation works as expected, but the feature count in QML still doesn't update until any button is clicked a second time.

tanneryould (approved):
LGTM and works as expected!

PR #1347, opened 2022-02-09
Gela/remove oauth

Gela:
We have a one-off sample called "OAuth Redirect" which attempted to showcase how to use an external browser + redirects vs embedding a browser in the app window. This was experimental at best, very clunky, and only worked on a few platforms. We are removing this to avoid confusion.


ArcGISRuntimeSDKQt_CppSamples/Features/FilterByDefinitionExpressionOrDisplayFilter/FilterByDefinitionExpressionOrDisplayFilter.cpp
```diff
@@ -133,3 +139,34 @@ void FilterByDefinitionExpressionOrDisplayFilter::queryFeatureCountInCurrentExte
 
   m_featureTable->queryFeatureCount(parameters);
 }
+
```
tanneryould, line 142: This looks like it was supposed to be included in a different PR?

tanneryould (approved):
Approved, but it looks like changes for "Filter by definition expression or display filter" were included in this PR.

ArcGISRuntimeSDKQt_CppSamples/Features/FilterByDefinitionExpressionOrDisplayFilter/FilterByDefinitionExpressionOrDisplayFilter.cpp
Gela: Yes, branching off an active branch on accident. Should be cleared up now

PR #1353, opened 2022-02-17
Fix "Apply Display Filter" button name

Gela:
Update button name to match the sample read me and the QML instance. This was identified during the verification process by Kathleen


tanneryould (approved):
LGTM

PR #1354, opened 2022-02-22
Tanneryould/contingent values

tanneryould:
This is a complicated API that I tried to clarify through a lot of in-line comments. Please especially review the comments and ensure they make sense. Thank you!


ArcGISRuntimeSDKQt_CppSamples/EditData/ContingentValues/ContingentValues.cpp
```diff
@@ -0,0 +1,297 @@
...
+  Basemap* fillmoreBasemap = new Basemap(fillmoreVTPK, this);
+  m_map = new Map(fillmoreBasemap, this);
+
+  // Load the geodatabase from a mobile geodatabase
+  const QString gdbDataPath = defaultDataPath() + "/ArcGIS/Runtime/Data/geodatabase/ContingentValuesBirdNests.geodatabase";
+  m_geodatabase = new Geodatabase(gdbDataPath, this);
+
+  // The coded value domains in this sample are hardcoded for simplicity, but can be retrieved from the GeodatabaseFeatureTable's Field's Domains.
+  m_statusValues = QVariantMap{{"Occupied", "OCCUPIED"}, {"Unoccupied", "UNOCCUPIED"}};
+  m_protectionValues = QVariantMap{{"Endangered", "ENDANGERED"}, {"Not endangered", "NOT_ENDANGERED"}, {"N/A", "NA"}};
```
Gela, line 82: Did you and Vivian come into an agreement about making this dynamic or keeping it hard coded?

ArcGISRuntimeSDKQt_CppSamples/EditData/ContingentValues/ContingentValues.qml
```diff
@@ -0,0 +1,210 @@
...
+            onWheel: wheel.accepted = attributePrompt.visible
+        }
+
+        Component.onCompleted: {
+            // Set and keep the focus on SceneView to enable keyboard navigation
+            forceActiveFocus();
+        }
+    }
+
+    // Declare the C++ instance which creates the scene etc. and supply the view
```
Gela, line 43:
```suggestion
    // Declare the C++ instance which creates the map etc. and supply the view
```

ArcGISRuntimeSDKQt_CppSamples/EditData/ContingentValues/ContingentValues.qml
```diff
@@ -0,0 +1,210 @@
...
+    ContingentValuesSample {
+        id: contingentValuesSample
+        mapView: view
+    }
+
+    Control {
+        id: attributePrompt
+
+        // Expand the attributes pane depending on where the user clicks or taps
```
Gela, line 52: The logic for the attributes pane might need to be reworked. Depending on where I click, the pane could be partly out of the view. But I cannot pan around to bring it back in view to complete the input fields. The only way I can interact with the map again is to hit "Discard". Below is an example. This is an issue in both C++ and QML

popup_construded

ArcGISRuntimeSDKQt_CppSamples/EditData/ContingentValues/ContingentValues.qml
```diff
@@ -0,0 +1,210 @@
...
+
+        visible: contingentValuesSample.featureAttributesPaneVisibe
+
+        contentItem: Column {
+            id: inputColumn
+            spacing: 5
+            padding: 10
+
+            Text {
+                text: "Status"
```
Gela, line 68: In the sample "Status" appears to be a mandatory field that needs a value before the new point can be saved. Why is there an empty field under Status dropdown?

ArcGISRuntimeSDKQt_CppSamples/EditData/ContingentValues/ContingentValues.qml
```diff
@@ -0,0 +1,210 @@
...
+
+                    contingentValuesSample.updateField("BufferSize", rangeValuesSpinBox.value);
+
+                    // Validate that all contingencies are valid, if so, enable the save button
+                    saveButton.enabled = contingentValuesSample.validateContingentValues();
+                }
+            }
+
+            Button {
+                id: saveButton
```
Gela, line 166: Should the Save button be grayed out until all relevant inputs are received? Right now if there is a missing input, the save button can be pressed but it doesn't seem to do anything which may be confusing to some.

ArcGISRuntimeSDKQt_CppSamples/EditData/ContingentValues/ContingentValues.qml
```diff
@@ -0,0 +1,210 @@
...
+                    const valid = contingentValuesSample.validateContingentValues();
+                    if (valid) {
+                        contingentValuesSample.createNewNest();
+                        contingentValuesSample.featureAttributesPaneVisibe = false;
+                    }
+                }
+            }
+
+            Button {
+                id: discardButton
```
Gela, line 186: Is it possible to also remove this pane when user clicks elsewhere in the map instead of just relying on clicking discard?

ArcGISRuntimeSDKQt_QMLSamples/EditData/ContingentValues/ContingentValues.qml
```diff
@@ -0,0 +1,374 @@
...
+                    id: gdb
+                    path: dataPath + "geodatabase/ContingentValuesBirdNests.geodatabase"
+
+                    onLoadStatusChanged: {
+                        if (loadStatus === Enums.LoadStatusLoaded) {
+                            birdNestsTable = gdb.geodatabaseFeatureTablesByTableName["BirdNests"];
+                            birdNestsTable.load();
+
+                            // Load the ContingentValuesDefinition to enable access to the GeodatabaseFeatureTable's ContingentValues data such as FieldGroups and ContingentValuesResults
+                            birdNestsTable.contingentValuesDefinition.load();
```
Gela, line 95: Upon opening the sample I get the following error referring to this line. Removing the call to load() does not seem to affect the sample.

> qrc:/Samples/EditData/ContingentValues/ContingentValues.qml:95: TypeError: Cannot call method 'load' of undefined

ArcGISRuntimeSDKQt_QMLSamples/EditData/ContingentValues/ContingentValues.qml
```diff
@@ -0,0 +1,374 @@
...
+            // Reset attribute panel values when the panel opens
+            statusComboBox.currentIndex = 0;
+            protectionComboBox.currentIndex = 0;
+            rangeValuesSpinBox.from = 0;
+            rangeValuesSpinBox.to = 0;
+        }
+    }
+
+    function getContingentValues(field, fieldGroup) {
+        const contingentValuesResult = birdNestsTable.contingentValues(newFeature, field);
```
Gela, line 298: When I select an option from the Status dropdown menu, I get an error regarding contingentValues here, complaining that it is not a function.

> qrc:/Samples/EditData/ContingentValues/ContingentValues.qml:298: TypeError: Property 'contingentValues' of object QmlGeodatabaseFeatureTable(0x28a546d8230) is not a function

ArcGISRuntimeSDKQt_QMLSamples/EditData/ContingentValues/ContingentValues.qml
```diff
@@ -0,0 +1,374 @@
...
+
+                    newFeature.attributes.replaceAttribute("BufferSize", rangeValuesSpinBox.value);
+
+                    // Validate that all contingencies are valid, if so, enable the save button
+                    saveButton.enabled = validateContingentValues();
+                }
+            }
+
+            Button {
+                id: saveButton
```
Gela, line 252: similar to the C++ sample, can this be greyed out until all input requirements are met?

ArcGISRuntimeSDKQt_QMLSamples/EditData/ContingentValues/ContingentValues.qml
```diff
@@ -0,0 +1,374 @@
...
+            Text {
+                text: "Protection"
+                font {
+                    bold: true
+                    pointSize: 11
+                }
+            }
+
+            ComboBox {
+                id: protectionComboBox
```
Gela, line 191: The protection field always appears to be unpopulated in the QML sample, whereas in the Cpp sample I have options for "endangered" and "not endangered"

Gela (changes requested):

ArcGISRuntimeSDKQt_QMLSamples/EditData/ContingentValues/ContingentValues.qml
tanneryould: My sample will crash if I don't call `contingentValuesDefinition.load()` which is expected behavior, weird. Well I moved that call to lower down and it shouldn't trigger until `birdNestsTable` loads.

ArcGISRuntimeSDKQt_QMLSamples/EditData/ContingentValues/ContingentValues.qml
tanneryould: That's potentially because the contingentValuesDefinition wasn't loading properly?

ArcGISRuntimeSDKQt_QMLSamples/EditData/ContingentValues/ContingentValues.qml
tanneryould: See C++ comment

ArcGISRuntimeSDKQt_QMLSamples/EditData/ContingentValues/ContingentValues.qml
tanneryould: Again, this is probably a result of the definition not loading properly

tanneryould (commented):
Thank you for your review Gela, hopefully I've addressed your feedback with my most recent PR

ArcGISRuntimeSDKQt_CppSamples/EditData/ContingentValues/ContingentValues.cpp
```diff
@@ -0,0 +1,297 @@
...
+    m_gdbFeatureTable->load();
+  });
+
+  connect(m_mapView, &MapQuickView::mouseClicked, this, &ContingentValues::createNewEmptyFeature);
+}
+
+// When the user clicks or taps on the map, instantiate a new feature and show the attribute form interface
+void ContingentValues::createNewEmptyFeature(QMouseEvent mouseEvent)
+{
+  if (m_newFeature)
```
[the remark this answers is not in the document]
Gela: I believe so, yeah. Please double check but I don't think it's necessary to set it to null first.

ArcGISRuntimeSDKQt_QMLSamples/EditData/ContingentValues/ContingentValues.qml
Gela: Yeah, seems to be working as expected now. Thanks

Gela:
Well done and kudos to you for taking this sample on. The previous issues seem to be addressed. I have one last comment, If I click on the map and open a pane, then continue to click around the map, it keeps adding points without any parameters.
As a user I would expect to see the attributes pane disappear after the first click, and on the second one a new one should pop up for a new data point. Hope I explained this clearly, but if not please let me know and I can elaborate.

tanneryould:
> Well done and kudos to you for taking this sample on. The previous issues seem to be addressed. I have one last comment, If I click on the map and open a pane, then continue to click around the map, it keeps adding points without any parameters. As a user I would expect to see the attributes pane disappear after the first click, and on the second one a new one should pop up for a new data point. Hope I explained this clearly, but if not please let me know and I can elaborate.

While I agree that clicking away on the map would be a good quality of life improvement, adding the code for that further obfuscates the Contingent Values API and bloats the app. I have set it so the MapView locks when the attribute pane is open and it will not re enable until the user clicks save or discard. I was able to get it to where the user could click on the map and have the attribute window close, but unfortunately if the user missed a button (and clicked somewhere on the rectangle instead) that would close the attribute prompt. Rather than creating a workaround, I'd rather ask that the user clicks "discard" to discard the new feature and close the attribute prompt.

Gela (approved):

PR #1361, opened 2022-03-17
Add geotriggers presentation demos

tanneryould:
These are the demos I wrote for the DevSummit 2022 geotriggers session. I'd like to share them as a further resource for our users.


Gela:
Are you intending on adding these to the sample viewer?

tanneryould:
@Gela no, I just intend to have them as a separate resource for our users. There have been a few cases where I'd like to share sample code that doesn't necessarily fit neatly into either C++/QML directory, so I'd like to create this `extras` directory.

Gela:
Does any other team do the same? I guess I'm a little hesitant since there's no design or documentation or a way to keep track of these so they don't go obsolete so they may be useless in a couple of years. Curious how others manage this

PR #1366, opened 2022-03-28
Delete FeatureLayerDefinitionexpression directory

Gela:
This sample was enhanced and renamed to FilterByDefinitionExpressionOrDisplayFilter. This instance is no longer valid and needs to be removed. All references to FeatureLayerDefinitionExpression were previously removed from the sample viewer.

issue: https://github.com/ArcGIS/afd-native-apis-doc/issues/2027


JamesMBallard (approved):
Is this also needed on v.next?

Gela:
@JamesMBallard Good catch, just created a PR to remove this from v.next too

PR #1371, opened 2022-04-18
Removing reference to internal value.

anmacdonald:
Identify KML sample reference internal API within Callout which is not publicly exposed. This caused issues with the callout rewrite, and in general is not what we should be demonstrating to users.

AFAIK this value was only there to deal with rounded corner offsets, and is not actually applicable to this use-case. Setting a hard-coded value equivalent to the same value.


ldanzinger (commented):
Ok, so the sample was incorrectly accessing an internal only property before... Makes sense.

Can you please update the QML sample as well?

ldanzinger:
@anmacdonald looks like Identify Raster Cell has same issue - https://github.com/Esri/arcgis-runtime-samples-qt/search?q=calloutFramePadding

anmacdonald:
@ldanzinger You're right. Updated.

ldanzinger (approved):

PR #1372, opened 2022-04-19
Fix DisplayFilter QML property

tanneryould:
The DisplayFilter object should have a `name` not an `id`. `id` is reserved and `filterId` is read-only.


tanneryould:
@Gela I'm not super familiar with this sample, but it looks like the `DisplayFilter`'s name in the [C++ implementation](https://github.com/Esri/arcgis-runtime-samples-qt/blob/v.next/ArcGISRuntimeSDKQt_CppSamples/Features/FilterByDefinitionExpressionOrDisplayFilter/FilterByDefinitionExpressionOrDisplayFilter.cpp#L139) is "Damaged Trees" and not "No Filter" like it is in QML. Similarly its `whereClause` is just `""` and not `"1=1"` like it is in QML. The `whereClause` doesn't seem to have an effect, but should we update either (or both) of these for parity between the two?

Gela:
Yes I agree

tanneryould:
@Gela I believe this is ready for review.
@GuillaumeBelz and @JamesMBallard, your review is welcomed but not necessary.

Thank you!

Gela (approved):
Oops, thought I approved it when I commented last night. LGTM

GuillaumeBelz (changes requested):
Some names don't follow the usages, can you rename `available_filters` and `display_filter_defintion`?

GuillaumeBelz:
@Gela Thank you, I didn't know.

GuillaumeBelz (approved):

PR #1375, opened 2022-04-21
Sample Viewer Source Code

tanneryould:
This PR will make the sample viewer source code public, allowing our users to build the sample viewer on their own in addition to giving the Runtime Qt samples team more control over it.


Gela (approved):
Since everything has already been reviewed in the other repo, this is good to go. Thanks

PR #1377, opened 2022-04-22
Tanneryould/update generate offline map paths

tanneryould:
The old path is wrong - an .mmpk is a zipped archive exported from pro, whereas the offline map task just writes a new directory. This PR removes that misleading ".mmpk" at the end of the directory path. It also cleans up the code a little by removing the unused toolkits and updating a deprecated property.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap/GenerateOfflineMap.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap/GenerateOfflineMap.cpp
@@ -105,7 +105,7 @@ void GenerateOfflineMap::generateMapByExtent(double xCorner1, double yCorner1, d
     if (generateJob)
     {
       // connect to the job's status changed signal
-      connect(generateJob, &GenerateOfflineMapJob::jobStatusChanged, this, [this, generateJob]()
+      connect(generateJob, &GenerateOfflineMapJob::statusChanged, this, [this, generateJob]()
       {
         // connect to the job's status changed signal to know once it is done
         switch (generateJob->jobStatus()) {
--- a/ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap/GenerateOfflineMap.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap/GenerateOfflineMap.cpp
@@ -99,7 +99,7 @@ void GenerateOfflineMap::generateMapByExtent(double xCorner1, double yCorner1, d
           this, [this](QUuid, const GenerateOfflineMapParameters& params)
   {
     // Take the map offline once the parameters are generated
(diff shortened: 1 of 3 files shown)


ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap/GenerateOfflineMap.cpp
```diff
@@ -99,13 +99,13 @@ void GenerateOfflineMap::generateMapByExtent(double xCorner1, double yCorner1, d
...
     // Take the map offline once the parameters are generated
-    GenerateOfflineMapJob* generateJob = m_offlineMapTask->generateOfflineMap(params, m_tempPath.path() + "/offlinemap.mmpk");
+    GenerateOfflineMapJob* generateJob = m_offlineMapTask->generateOfflineMap(params, m_tempPath.path() + "/offlinemap");
 
     // check if there is a valid job
     if (generateJob)
     {
       // connect to the job's status changed signal
-      connect(generateJob, &GenerateOfflineMapJob::jobStatusChanged, this, [this, generateJob]()
+      connect(generateJob, &GenerateOfflineMapJob::statusChanged, this, [this, generateJob]()
```
lsmallwood, line 108: this new signal actually gives you the `status` as a signal argument. We should use this rather than checking the job property again since it avoids a timing issue when the status has moved on since the signal was fired.

```suggestion
      connect(generateJob, &GenerateOfflineMapJob::statusChanged, this, [this](JobStatus jobStatus)
```

lsmallwood (approved):
LGTM @tanneryould - thanks for fixing that up 👍

ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap/GenerateOfflineMap.cpp
tanneryould: @lsmallwood I think we still need to capture `generateJob` right? `JobStatus` only provides access to the status enum. Or should I make `generateJob` a member variable?

ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap/GenerateOfflineMap.cpp
```diff
@@ -105,10 +105,10 @@ void GenerateOfflineMap::generateMapByExtent(double xCorner1, double yCorner1, d
     if (generateJob)
     {
       // connect to the job's status changed signal
-      connect(generateJob, &GenerateOfflineMapJob::statusChanged, this, [this, generateJob]()
+      connect(generateJob, &GenerateOfflineMapJob::statusChanged, this, [this, generateJob](JobStatus jobStatus)
```
lsmallwood, line 108: don't need to capture `generateJob` anymore

lsmallwood: oh, sorry my bad. you do use the job further down. that's fine then. you could move all of that completion stuff out to the `jobDone` signal but this is fine.

lsmallwood (approved):

PR #1383, opened 2022-04-26
Update jobStatusChanged signal to statusChanged

Gela:
Update deprecated "jobStatusChanged" signal to statusChanged.

```issue: common-samples/3494```

### DIFF
--- a/ArcGISRuntimeSDKQt_QMLSamples/Maps/ApplyScheduledMapUpdates/ApplyScheduledMapUpdates.qml
+++ b/ArcGISRuntimeSDKQt_QMLSamples/Maps/ApplyScheduledMapUpdates/ApplyScheduledMapUpdates.qml
@@ -133,7 +133,7 @@ Rectangle {
                 syncJob = offlineMapSyncTask.syncOfflineMap(createDefaultOfflineMapSyncParametersResult);
 
                 // connect to the job signals
-                syncJob.jobStatusChanged.connect(()=> {
+                syncJob.statusChanged.connect(()=> {
                                                      if (syncJob.jobStatus === Enums.JobStatusSucceeded) {
                                                          const mapResult = syncJob.result;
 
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

**Review by tanneryould — CHANGES_REQUESTED** [review]
`statusChanged` provides access to the `JobStatus` class which we should use in lieu of `job->jobStatus()` . See [Generate offline map](https://github.com/Esri/arcgis-runtime-samples-qt/blob/v.next/ArcGISRuntimeSDKQt_CppSamples/Maps/GenerateOfflineMap/GenerateOfflineMap.cpp#L108) for an example.

**Comment by Gela** [comment]
@lsmallwood could you take a look and verify it's good to be merged?

File: ArcGISRuntimeSDKQt_QMLSamples/Maps/ApplyScheduledMapUpdates/ApplyScheduledMapUpdates.qml [hunk 1]
```diff
@@ -133,7 +133,7 @@ Rectangle {
                 syncJob = offlineMapSyncTask.syncOfflineMap(createDefaultOfflineMapSyncParametersResult);
 
                 // connect to the job signals
-                syncJob.jobStatusChanged.connect(()=> {
+                syncJob.statusChanged.connect(()=> {
```
lsmallwood, line 136: i think you should be able to get the status as an arg to this connection, rather than having to poll the `syncJob` again

ArcGISRuntimeSDKQt_QMLSamples/Maps/DownloadPreplannedMap/DownloadPreplannedMap.qml
```diff
@@ -138,7 +138,7 @@ Rectangle {
                 } else {
                     job = offlineMapTask.downloadPreplannedOfflineMapWithParameters(createDefaultDownloadPreplannedOfflineMapParametersResult, path);
 
-                    job.jobStatusChanged.connect(()=> {
+                    job.statusChanged.connect(()=> {
```
lsmallwood, line 141: same here - can you get the status as an arg to the connection?

lsmallwood (approved):
Looks good @Gela. I think we can avoid polling the job again for the status in the qml samples but otherwise looks good.

Did we have any connections for `messagesChanged`? That can be replaced as well if so.

tanneryould (approved):

PR #1386, opened 2022-05-02
Move FeatureLayer declaration in Qml DisplayLayerViewDrawState

AndrewBladon:
This PR moves the FeatureLayer declared in the Qml `DisplayLayerViewDrawState` sample out of the scope of the Map declaration.

This change is required following `qt-common` issue [7206](https://devtopia.esri.com/runtime/qt-common/issues/7206) (PR: [2660](https://devtopia.esri.com/runtime/qt/pull/2660)), which has changed the way that nested feature layers are loaded. This resulted in the behaviour of this sample changing, so the FeatureLayer declaration has been moved to ensure the sample runs as expected.


tanneryould (approved):
LGTM, thank you!

PR #1387, opened 2022-05-03
Merge commits from main to v.next

Gela:
I looked through the commit diff list [here ](https://github.com/Esri/arcgis-runtime-samples-qt/compare/v.next...main)and cherry-picked the commits over from main to v.next. A good number of those commits would come back as "empty" after trying to cherry-pick them which means that they might've already been copied over with a different pull request. Please let me know if I missed any other changes.


tanneryould (approved):
LGTM thank you!

PR #1391, opened 2022-05-17
Updating Callout.

anmacdonald:
Covers: https://github.com/Esri/arcgis-runtime-samples-qt/issues/1390

Deprecating usage of `calloutContent` and simplifying the sample to not have to deal with custom sizing logic using bindings.


tanneryould (approved):
LGTM and this looks like a great improvement

PR #1394, opened 2022-06-01
[New Sample] Alternate symbols with class break renderer

Gela:


Gela:
New PR [here](https://github.com/Esri/arcgis-runtime-samples-qt/pull/1396). Closing this one since both samples are using the wrong template.

PR #1405, opened 2022-06-15
Add telemetry

tanneryould:
# Description

Adds Google Analytics telemetry to the sample viewer. Requires coordination with DevOps to implement into deployed Sample Viewer. This is done with a `GAnalytics` singleton that can be invoked from any .qml file that has `import Telemetry 0.1`. The class sends events via HTTP Post to Google Analytics.

## Type of change

- [ ] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

## Checklist

- [ ] Runs and compiles on all active platforms
- [x] Self-review of changes
- [x] No unrelated changes have been made to any other code or project files
- [x] Code is commented with correct formatting (CTRL+i)

### DIFF
--- a/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
+++ b/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
@@ -1,0 +1,177 @@
+#include "GAnalytics.h"
+
+#include <QDebug>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
+#include <QOperatingSystemVersion>
+#include <QQmlEngine>
+#include <QSettings>
+#include <QUuid>
+
+#ifdef GANALYTICS_API_KEY
+#ifdef GANALYTICS_STREAM_ID
+GAnalytics::GAnalytics(QObject* parent /* = nullptr */):
+  QObject(parent),
+  m_apiSecret(QUOTE(GANALYTICS_API_KEY)),
+  m_measurementId(QUOTE(GANALYTICS_STREAM_ID))
+{
+}
+#endif // GANALYTICS_STREAM_ID
+#endif // GANALYTICS_API_KEY
+
+GAnalytics::GAnalytics(QObject* parent /* = nullptr */):
+  QObject(parent),
+  m_apiSecret(""),
+  m_measurementId("")
+{
+}
+
+GAnalytics::~GAnalytics()
+{
+  endSession();
+}
+
+void GAnalytics::init()
+{
+  if (m_apiSecret == "" || m_measurementId == "")
+  {
+    m_telemetryEnabled = false;
+    setIsVisible(false);
+    return;
+  }
+
+  QSettings settings;
+  generateClientId();
+  if (!settings.contains("GAnalytics-telemetry-enabled"))
+  {
+    m_telemetryEnabled = true;
+    settings.setValue("GAnalytics-telemetry-enabled", m_telemetryEnabled);
+    setIsVisible(true);
+  }
+  else
+  {
+    m_telemetryEnabled = settings.value("GAnalytics-telemetry-enabled").toBool();
+  }
+}
+
+void GAnalytics::postEvent(const QString eventName, QVariantMap parameters)
+{
+  if (!m_telemetryEnabled || m_apiSecret == "" || m_measurementId == "")
+    return;
+
+#ifdef CPP_VIEWER
+  parameters.insert("sample_viewer_type", "Cpp");
+#else // QML_VIEWER
+  parameters.insert("sample_viewer_type", "QML");
+#endif // CPP_VIEWER / QML_VIEWER
+
+  auto os = QOperatingSystemVersion::current();
+
+  parameters.insert("operating_system", os.name() + " " + QString::number(os.majorVersion()) + "." + QString::number(os.minorVersion()));
+
+  int sessionTime = (QDateTime::currentSecsSinceEpoch()-m_startTime)*1000;
+
+  // Google will not accept the POST request if the engagement time is 0 or undefined
+  if (sessionTime <= 0)
+    sessionTime = 1;
+
+  parameters.insert("engagement_time_msec", sessionTime);
+
+  QJsonObject event;
+  event.insert("name", eventName);
+  event.insert("params", QJsonObject::fromVariantMap(parameters));
+
+  QJsonObject body;
+  body.insert("client_id", m_clientId);
+  body.insert("events", event);
+  QNetworkAccessManager* mgr = new QNetworkAccessManager();
+  const QUrl url(QString("https://google-analytics.com/mp/collect?"
+                         "api_secret=" + m_apiSecret +
+                         "&measurement_id=" + m_measurementId));
+
+  QNetworkRequest request(url);
+  request.setHeader(QNetworkRequest::ContentTypeHeader, "application, json");
+  QJsonDocument data(body);
+
+  mgr->post(request, data.toJson());
+}
+
+void GAnalytics::startSession()
+{
+  m_startTime = QDateTime::currentSecsSinceEpoch();
+  postEvent("open_sample_viewer", QVariantMap());
+}
+
+void GAnalytics::endSession()
+{
+  QVariantMap params;
+  params.insert("session_duration_sec", QDateTime::currentSecsSinceEpoch() - m_startTime);
+  postEvent("close_sample_viewer", params);
+}
+
+void GAnalytics::generateClientId()
+{
+  QSettings settings;
+  if (!settings.contains("GAnalytics-clientId"))
+  {
+    m_clientId= QUuid::createUuid().toString();
+    settings.setValue("GAnalytics-clientId", m_clientId);
+  }
+  else
+  {
+    m_clientId = settings.value("GAnalytics-clientId").toString();
+  }
+}
+
+QString GAnalytics::apiSecret() const
+{
+  return m_apiSecret;
+}
+
+void GAnalytics::setApiSecret(const QString apiSecret)
+{
+  m_apiSecret = apiSecret;
+}
+
+QString GAnalytics::measurementId() const
+{
+  return m_measurementId;
+}
+
+void GAnalytics::setMeasurementId(const QString measurementId)
+{
+  m_measurementId = measurementId;
+}
+
+bool GAnalytics::telemetryEnabled() const
+{
+  return m_telemetryEnabled;
+}
+
+void GAnalytics::setTelemetryEnabled(const bool telemetryEnabled)
+{
+  m_telemetryEnabled = telemetryEnabled;
+
+  QSettings settings;
+  settings.setValue("GAnalytics-telemetry-enabled", m_telemetryEnabled);
+
+  if (telemetryEnabled)
+    startSession();
+}
+
+bool GAnalytics::isVisible() const
+{
+  return m_isVisible;
+}
+
+void GAnalytics::setIsVisible(bool isVisible)
+{
+  m_isVisible = isVisible;
+}
+
+QString GAnalytics::clientId() const
+{
+  return m_clientId;
+}
--- a/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.h
+++ b/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.h
@@ -1,0 +1,53 @@
+#ifndef GANALYTICS_H
+#define GANALYTICS_H
+
+#include <QDateTime>
+#include <QObject>
+#include <QVariantMap>
+
+class GAnalytics : public QObject
+{
+  Q_OBJECT
+
+  Q_PROPERTY(QString clientId READ clientId CONSTANT)
+  Q_PROPERTY(QString apiSecret READ apiSecret WRITE setApiSecret NOTIFY apiSecretChanged)
+  Q_PROPERTY(QString measurementId READ measurementId WRITE setMeasurementId NOTIFY measurementIdChanged)
+  Q_PROPERTY(bool telemetryEnabled READ telemetryEnabled WRITE setTelemetryEnabled NOTIFY telemetryEnabledChanged)
+  Q_PROPERTY(bool isVisible READ isVisible WRITE setIsVisible NOTIFY isVisibleChanged)
+
+public:
+  explicit GAnalytics(QObject* parent = nullptr);
+
+  ~GAnalytics() override;
+
+  Q_INVOKABLE virtual void init();
+  Q_INVOKABLE void postEvent(const QString eventName, QVariantMap parameters);
+
+signals:
+  void apiSecretChanged();
+  void measurementIdChanged();
+  void telemetryEnabledChanged();
+  void isVisibleChanged();
+
+private:
+  void setApiSecret(const QString apiSecret);
+  void setMeasurementId(const QString measurementId);
+  void setTelemetryEnabled(const bool telemetryEnabled);
+  void setIsVisible(const bool isVisible);
+  void startSession();
+  void endSession();
+  void generateClientId();
+  QString apiSecret() const;
+  QString clientId() const;
+  QString measurementId() const;
+  bool telemetryEnabled() const;
+  bool isVisible() const;
+  QString m_apiSecret;
+  QString m_measurementId;
+  QString m_clientId;
+  qint64 m_startTime;
+  bool m_telemetryEnabled;
+  bool m_isVisible;
+};
+
+#endif // GANALYTICS_H
--- a/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/mainSample.cpp
+++ b/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/mainSample.cpp
@@ -26,6 +26,8 @@
 #include <QQuickWindow>
 #include <QtGlobal>
 
+#include "GAnalytics.h"
+
 #ifdef QT_WEBVIEW_WEBENGINE_BACKEND
 #  include <QtWebEngine>
 #endif // QT_WEBVIEW_WEBENGINE_BACKEND
@@ -56,7 +58,7 @@
 #include "DrawOrderLayerListModel.h"
 
 #ifdef ESRI_BUILD
-#include "../../../../buildnum/buildnum.h"
+#include "../../../../../buildnum/buildnum.h"
 #endif
 
 // All CPP Samples
@@ -293,6 +295,7 @@
 #define QUOTE(x) STRINGIZE(x)
 
 QObject* esriSampleManagerProvider(QQmlEngine* engine, QJSEngine* scriptEngine);
+QObject* gAnalyticsProvider(QQmlEngine* engine, QJSEngine* scriptEngine);
 QObject* syntaxHighlighterProvider(QQmlEngine* engine, QJSEngine* scriptEngine);
 void registerClasses();
 void registerCppSampleClasses();
@@ -595,6 +598,8 @@ void registerClasses()
 
   qmlRegisterSingletonType<DownloadSampleManager>("Esri.ArcGISRuntimeSamples", 1, 0, "SampleManager",
                                                   &esriSampleManagerProvider);
+  // Register the Google Analytics class
+  qmlRegisterSingletonType<GAnalytics>("Telemetry", 0, 1, "GAnalytics", &gAnalyticsProvider);
 
   qmlRegisterSingletonType<SyntaxHighlighter>("Esri.ArcGISRuntimeSamples", 1, 0, "SyntaxHighlighter",
                                               &syntaxHighlighterProvider);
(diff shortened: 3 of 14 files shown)


ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
```diff
@@ -0,0 +1,177 @@
...
+#endif // CPP_VIEWER / QML_VIEWER
+
+  auto os = QOperatingSystemVersion::current();
+
+  parameters.insert("operating_system", os.name() + " " + QString::number(os.majorVersion()) + "." + QString::number(os.minorVersion()));
+
+  int sessionTime = (QDateTime::currentSecsSinceEpoch()-m_startTime)*1000;
+
+  // Google will not accept the POST request if the engagement time is 0 or undefined
+  if (sessionTime <= 0)
```
Gela, line 77: Why do we set the engagement time to 1 here rather than discarding it altogether?

Gela (approved):

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
```diff
@@ -0,0 +1,177 @@
...
+  parameters.insert("sample_viewer_type", "Cpp");
+#else // QML_VIEWER
+  parameters.insert("sample_viewer_type", "QML");
+#endif // CPP_VIEWER / QML_VIEWER
+
+  auto os = QOperatingSystemVersion::current();
+
+  parameters.insert("operating_system", os.name() + " " + QString::number(os.majorVersion()) + "." + QString::number(os.minorVersion()));
+
+  int sessionTime = (QDateTime::currentSecsSinceEpoch()-m_startTime)*1000;
```
ldanzinger, line 74: 🙃 
```suggestion
  int sessionTime = (QDateTime::currentSecsSinceEpoch() - m_startTime) * 1000;
```

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+  QJsonObject event;
+  event.insert("name", eventName);
+  event.insert("params", QJsonObject::fromVariantMap(parameters));
+
+  QJsonObject body;
+  body.insert("client_id", m_clientId);
+  body.insert("events", event);
+  QNetworkAccessManager* mgr = new QNetworkAccessManager();
+  const QUrl url(QString("https://google-analytics.com/mp/collect?"
```
ldanzinger, line 90: 🙃 Consider `QString.arg` to make it a bit easier to maintain vs concatenated strings

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.h
```diff
@@ -0,0 +1,53 @@
...
+#include <QDateTime>
+#include <QObject>
+#include <QVariantMap>
+
+class GAnalytics : public QObject
+{
+  Q_OBJECT
+
+  Q_PROPERTY(QString clientId READ clientId CONSTANT)
+  Q_PROPERTY(QString apiSecret READ apiSecret WRITE setApiSecret NOTIFY apiSecretChanged)
```
ldanzinger, line 13: 🙃 For some of these boiler plate properties where you are just getting and setting the member, you can use the MEMBER keyword and remove the getters/setters all together

```suggestion
  Q_PROPERTY(QString apiSecret MEMBER m_apiSecret NOTIFY apiSecretChanged)
```

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.h
```diff
@@ -0,0 +1,53 @@
...
+  Q_PROPERTY(bool telemetryEnabled READ telemetryEnabled WRITE setTelemetryEnabled NOTIFY telemetryEnabledChanged)
+  Q_PROPERTY(bool isVisible READ isVisible WRITE setIsVisible NOTIFY isVisibleChanged)
+
+public:
+  explicit GAnalytics(QObject* parent = nullptr);
+
+  ~GAnalytics() override;
+
+  Q_INVOKABLE virtual void init();
+  Q_INVOKABLE void postEvent(const QString eventName, QVariantMap parameters);
```
ldanzinger, line 24: 🔧 should be `&`

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.h
```diff
@@ -0,0 +1,53 @@
...
+
+signals:
+  void apiSecretChanged();
+  void measurementIdChanged();
+  void telemetryEnabledChanged();
+  void isVisibleChanged();
+
+private:
+  void setApiSecret(const QString apiSecret);
+  void setMeasurementId(const QString measurementId);
```
ldanzinger, line 34: 🔧 
```suggestion  
  void setApiSecret(const QString& apiSecret);
  void setMeasurementId(const QString& measurementId);
```

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/mainSample.cpp
```diff
@@ -56,7 +58,7 @@
 #include "DrawOrderLayerListModel.h"
 
 #ifdef ESRI_BUILD
-#include "../../../../buildnum/buildnum.h"
+#include "../../../../../buildnum/buildnum.h"
```
ldanzinger, line 61: ❓ Is this change related to this PR or to getting the builds switched over to the public repo? If the latter, better to keep it separate

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/mainSample.cpp
```diff
@@ -595,6 +598,8 @@ void registerClasses()
 
   qmlRegisterSingletonType<DownloadSampleManager>("Esri.ArcGISRuntimeSamples", 1, 0, "SampleManager",
                                                   &esriSampleManagerProvider);
+  // Register the Google Analytics class
+  qmlRegisterSingletonType<GAnalytics>("Telemetry", 0, 1, "GAnalytics", &gAnalyticsProvider);
```
ldanzinger, line 602: ❓ Why is the version 0.1? I would've thought this would be 1.0 unless it is being considered as beta/alpha/tech preview or something along those lines. As far as I can tell, it is fully functional and is the 1.0 version.

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/CategoryCard.qml
```diff
@@ -95,6 +96,7 @@ Component {
             sampleListView.currentCategory = displayName;
             SampleManager.currentCategory = SampleManager.categories.get(index);
             stackView.push(sampleListView);
+            GAnalytics.postEvent("category_selected", {"category name": displayName});
```
ldanzinger, line 99: 🔧 Here are some more cases of magic strings that we should consider centralizing in some way. There are other cases in the other QML files as well

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/CategoryGridView.qml
```diff
@@ -49,6 +50,10 @@ Page {
             font.pixelSize: 14
             placeholderText: qsTr("Find a sample...")
             padding: 10
+
+            onTextChanged: {
+                searchQueryEventTimer.start();
```
ldanzinger, line 55: ❓ I wonder if there is a better way to globally enable/disable this code from running? Currently the check is in post event in C++ and the network calls won't go out if it is disabled. However, all of this overhead is still needlessly running when disabled

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/GAnalyticsView.qml
```diff
@@ -0,0 +1,92 @@
+import QtQuick 2.0
+import Telemetry 0.1
+import QtGraphicalEffects 1.0
```
ldanzinger, line 3: 🤡 No action needed at this point since this is already used elsewhere in the samples, but worth pointing out that QtGraphicalEffects isn't in Qt 6, so this will need to change at that point

ldanzinger (changes requested):
@tanneryould this is looking excellent. Nice work getting it going so quickly. I've got some comments that'll hopefully make this easier to maintain long term as well as streamline performance a bit. I'd also make sure that we get this tested on all platforms thoroughly to ensure QSettings and some of the OS info works as expected

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
tanneryould: Perhaps I should be a little clearer in this comment, but the session time is a required parameter
// Google will not accept the POST request if the engagement time is 0 or undefined

tanneryould:
@JamesMBallard and @GuillaumeBelz I'm inviting you to review these changes if you want because I'm sure you have valuable feedback to share, but neither of you should feel obligated to do so. All the same, thank you!

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
```diff
@@ -0,0 +1,177 @@
+#include "GAnalytics.h"
```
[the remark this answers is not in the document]
JamesMBallard: Are these adapted or forked from some other repo? If so we need to honor those copyrights. If not, disregard.

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
```diff
@@ -0,0 +1,164 @@
...
+{
+}
+#endif // GANALYTICS_API_KEY && GANALYTICS_STREAM_ID
+
+GAnalytics::GAnalytics(QObject* parent /* = nullptr */):
+  QObject(parent),
+  m_apiSecret(""),
+  m_measurementId("")
+{
+}
```
JamesMBallard, line 42: This does not look correct. If you hit the first ifdef, then you will compile two matching constructors. I'd recommend something like only ifdef'ing the members you want to be different:
```suggestion
GAnalytics::GAnalytics(QObject* parent /* = nullptr */):
  QObject(parent),
#if defined (GANALYTICS_API_KEY) && defined (GANALYTICS_STREAM_ID)
  m_apiSecret(QUOTE(GANALYTICS_API_KEY)),
  m_measurementId(QUOTE(GANALYTICS_STREAM_ID))
#else
  m_apiSecret(""),
  m_measurementId("")
#endif // GANALYTICS_API_KEY && GANALYTICS_STREAM_ID
{
}
```

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+GAnalytics::~GAnalytics()
+{
+  endSession();
+}
+
+void GAnalytics::init()
+{
+  if (m_apiSecret == "" || m_measurementId == "")
```
JamesMBallard, line 51: Use canonical QString isEmpty.
```suggestion
  if (m_apiSecret.isEmpty() || m_measurementId.isEmpty())
```

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.h
```diff
@@ -0,0 +1,65 @@
...
+  void setTelemetryEnabled(const bool telemetryEnabled);
+  void setIsVisible(const bool isVisible);
+  void startSession();
+  void endSession();
+  void generateClientId();
+  bool telemetryEnabled() const;
+  bool isVisible() const;
+
+  QString m_apiSecret;
+  QString m_measurementId;
```
JamesMBallard, line 54: Since these are only assigned in the initializer list and never modified, they can be const.

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
```diff
@@ -0,0 +1,164 @@
...
+  m_defaultParameters.insert("sample_viewer_type", "QML");
+#endif // CPP_VIEWER / QML_VIEWER
+
+  auto os = QOperatingSystemVersion::current();
+  m_defaultParameters.insert("operating_system", QString{"%1 %2.%3"}.arg(os.name(), os.majorVersion(), os.minorVersion()));
+}
+
+void GAnalytics::postEvent(const QString& eventName, QVariantMap parameters)
+{
+  if (!m_telemetryEnabled || m_apiSecret == "" || m_measurementId == "")
```
JamesMBallard, line 87: Prefer `isEmpty` to empty string compares.

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+  m_networkAccessManager = new QNetworkAccessManager();
+
+#ifdef CPP_VIEWER
+  m_defaultParameters.insert("sample_viewer_type", "Cpp");
+#else // QML_VIEWER
+  m_defaultParameters.insert("sample_viewer_type", "QML");
+#endif // CPP_VIEWER / QML_VIEWER
+
+  auto os = QOperatingSystemVersion::current();
```
JamesMBallard, line 81: Can this be const?

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
```diff
@@ -0,0 +1,164 @@
...
+  m_defaultParameters.insert("operating_system", QString{"%1 %2.%3"}.arg(os.name(), os.majorVersion(), os.minorVersion()));
+}
+
+void GAnalytics::postEvent(const QString& eventName, QVariantMap parameters)
+{
+  if (!m_telemetryEnabled || m_apiSecret == "" || m_measurementId == "")
+    return;
+
+  int sessionTime = (QDateTime::currentSecsSinceEpoch() - m_startTime) * 1000;
```
JamesMBallard, line 90: Any worry about overflowing sessionTime? Probably not, looks like seconds. Potentially a different name here would make that clearer.

```suggestion
  int sessionTimeInSeconds = (QDateTime::currentSecsSinceEpoch() - m_startTime) * 1000;
```

Optional, since I'm not sure if this is correct.

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
```diff
@@ -0,0 +1,164 @@
...
+  event.insert("params", QJsonObject::fromVariantMap(parameters));
+
+  QJsonObject body;
+  body.insert("client_id", m_clientId);
+  body.insert("events", event);
+  const QUrl url(QString{"https://google-analytics.com/mp/collect?api_secret=%1&measurement_id=%2"}.arg(m_apiSecret, m_measurementId));
+
+  QNetworkRequest request(url);
+  request.setHeader(QNetworkRequest::ContentTypeHeader, "application, json");
+  QJsonDocument data(body);
```
JamesMBallard, line 110: Can this be const?

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+  m_networkAccessManager->post(request, data.toJson());
```
JamesMBallard, line 112: Do you need to check for the response, or any errors? This seems suspicious to fire and forget here.

Disregard if you have alternate logic that doesn't need the QNetworkReply.

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/CategoryGridView.qml
```diff
@@ -88,4 +94,15 @@ Page {
...
     // hidden. This is an issue on iOS and Android devices.
     onVisibleChanged:  searchBar.focus = false;
+
+    Timer {
+        id: searchQueryEventTimer
+        interval: 2500 // milliseconds
+        running: false
+        repeat: false
+        onTriggered: {
+            if (searchBar.text !== "" && GAnalytics.telemetryEnabled)
```
JamesMBallard, line 104: I don't know if this makes any real difference, but a bool check should be faster than a string check. I would write the bool check first so it can short circuit and avoid any string check at all if telemetry is disabled. You need to check both anyways.

Actually, another idea is to allow the API to be used with impunity but the calls all become no-ops if telemetry is disabled. That moves the responsibility of managing state exclusively in the C++ implementation. This could be a big refactor and I don't know if there is scope for that.

If you do that, code like this can simply always call `GAnalytics.postEvent`, but that logic determines if anything should happen or not.

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/SampleListView.qml
```diff
@@ -101,6 +102,7 @@ Page {
                             drawer.close();
                             // launch sample...
                             SampleManager.currentSample = sample;
+                            GAnalytics.postEvent("sample_opened", {"sample name": sample.name, "referrer": "category list"});
```
JamesMBallard, line 105: Should this also be behind an `if (GAnalytics.telemetryEnabled)` check? My previous recommendation would cover that (move state logic to the class itself).

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/SearchView.qml
```diff
@@ -52,6 +53,7 @@ ListView {
                     drawer.close();
                     // launch sample...
                     SampleManager.currentSample = sample;
+                    GAnalytics.postEvent("sample_opened", {"sample name": sample.name, "referrer": "search"});
```
JamesMBallard, line 56: Ditto does this need a check? `if (GAnalytics.telemetryEnabled)` ?

There are a lot of these, so I won't keep asking.

JamesMBallard (commented):
Very cool. I added some comments.

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/GAnalytics.cpp
tanneryould: No need to check for a response because we wouldn't do anything with it. This is intended to work silently in the background. Furthermore Google Analytics doesn't return anything from the request apart from Status 202.

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/CategoryGridView.qml
tanneryould: I have it set up so `postEvent` only runs if `telemetryEnabled == false`, so it's handled on the C++ side rather than needing to remember to check in QML.

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/SampleListView.qml
tanneryould: Yeah the state logic is built into the `postEvent` method

JamesMBallard (approved):

ldanzinger (approved):
LGTM - I recommend logging a follow up issue to split out the embedded strings. I think it will make it more maintainable in the long run

PR #1411, opened 2022-06-21
Recognize Arcade as a proper noun

Gela:
# Description

This PR accounts for making all references to 'Arcade' capitalized in our sample read me and metadata, since Esri's creation of this expression language as a proper noun. For instance, you can see all written usages of it capitalized here: https://developers.arcgis.com/arcade/

## Type of change

- [ ] Bug fix
- [x] Other enhancement


tanneryould (approved):
Url needs to be lower case, but no need for re-review.

PR #1441, opened 2022-08-04
Tanner/ios qstandardpath

tanneryould:
When I was building the Qt C++ sample viewer for iOS I needed to resolve these import issues before I could successfully build. Most of the files that reference `QStandardPaths` already have this include, but this PR adds the include statement to the remaining files. Adding this line fixed the build errors.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
@@ -32,6 +32,7 @@
 
 #include <QDir>
 #include <algorithm>
+#include <QStandardPaths>
 
 using namespace Esri::ArcGISRuntime;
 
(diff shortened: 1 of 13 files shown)


ArcGISRuntimeSDKQt_CppSamples/Layers/TileCacheLayer/TileCacheLayer.cpp
```diff
@@ -26,6 +26,7 @@
 #include "TileCache.h"
 
 #include <QDir>
+#include <QStandardPaths>
```
JamesMBallard, line 29: Ideally this would be in an `Q_OS_IOS` ifdef, but you could make the same argument for QDir as well (only for Android).

This is good to keep the code tidy.

ArcGISRuntimeSDKQt_CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -32,6 +32,7 @@
 
 #include <QDir>
 #include <algorithm>
+#include <QStandardPaths>
```
JamesMBallard, line 35: I recommend keeping these alphabetical. The Qt team's git hooks will do this automatically if you have them installed.

This recommendation applies throughout.

JamesMBallard (approved):

PR #1451, opened 2022-08-19
Tanner/navigate with reroute qml

tanneryould:
# Description

Adds the QML implementation of the rerouting sample.

I moved components around a bit to hopefully improve readability (the routing sample was written 3 years ago and we've changed some standards since then), so unfortunately it looks like there were a lot more changes than are actually present. The UI is now at the end of the file, RouteTracker/RouteTask are no longer a part of map view, I removed some global variables and instead am getting the properties directly from the components so it's clear where the parameters are coming from, etc.

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS


Gela (approved):
LGTM, thanks for taking this on!

PR #1453, opened 2022-08-24
nearest vertex sample: fix panning issue

Gela:
# Description

Remove extra property that prevents users from panning. 

## Type of change

- [x] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [x] Windows
- [ ] Android
- [ ] Linux
- [ ] macOS
- [ ] iOS

### DIFF
--- a/ArcGISRuntimeSDKQt_QMLSamples/Geometry/NearestVertex/NearestVertex.qml
+++ b/ArcGISRuntimeSDKQt_QMLSamples/Geometry/NearestVertex/NearestVertex.qml
@@ -120,7 +120,6 @@ Rectangle {
     }
 
     MouseArea {
-        anchors.fill: parent
         onClicked: {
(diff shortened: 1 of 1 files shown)


ArcGISRuntimeSDKQt_QMLSamples/Geometry/NearestVertex/NearestVertex.qml
```diff
@@ -120,7 +120,6 @@ Rectangle {
     }
 
     MouseArea {
-        anchors.fill: parent
```
ldanzinger, line 123: this allows you to pan, but does the onClicked signal ever get called now? Presumably, this all should be be moved into a onMouseClicked signal handler on MapView instead of setting up a separate MouseArea

ldanzinger (approved):
LGTM

PR #1461, opened 2022-09-01
Replace QStringRef with QStringView

Gela:
# Description

Required for moving to Qt6. Per [Qt documentation](https://doc.qt.io/qt-6/qstringview.html) const ref to QStringView works but it is slower so I replaced that with just using a QStringView.

```c++
void myfun2(const QStringView &sv); // compiles and works, but slower
```

## Type of change

- [ ] Bug fix
- [ ] Other enhancement

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/GODictionaryRenderer/GODictionaryRenderer.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/GODictionaryRenderer/GODictionaryRenderer.cpp
@@ -190,7 +190,7 @@ void GODictionaryRenderer::parseXmlFile()
         if (!currentElementName.isEmpty())
         {
           // Get the text and store it as the current element's value
-          QStringRef trimmedText = m_xmlParser.text().trimmed();
+          QStringView trimmedText = m_xmlParser.text().trimmed();
           if (!trimmedText.isEmpty())
           {
             elementValues[currentElementName] = trimmedText.toString();
--- a/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/GODictionaryRenderer/GODictionaryRenderer.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/GODictionaryRenderer/GODictionaryRenderer.cpp
@@ -190,7 +190,7 @@ void GODictionaryRenderer::parseXmlFile()
         if (!currentElementName.isEmpty())
         {
           // Get the text and store it as the current element's value
-          QStringView trimmedText = m_xmlParser.text().trimmed();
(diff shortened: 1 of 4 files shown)


ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/GODictionaryRenderer/GODictionaryRenderer.cpp
```diff
@@ -190,7 +190,7 @@ void GODictionaryRenderer::parseXmlFile()
         if (!currentElementName.isEmpty())
         {
           // Get the text and store it as the current element's value
-          QStringRef trimmedText = m_xmlParser.text().trimmed();
+          QStringView trimmedText = m_xmlParser.text().trimmed();
```
GuillaumeBelz, line 193: Is there a guideline in samples about using `auto`? Here, `QXmlReader::text` returns `QStringView`, then it's possible to use `auto`.

And maybe `trimmedText` can be `const`?

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/SearchFilterKeywordCriteria.cpp
```diff
@@ -35,20 +35,20 @@ QString highlightedMidpoint(const QString& string,
 
     // Find the portion of string that goes before the catchment.
     int startPos = std::max(0, pos-catchment);
```
GuillaumeBelz, line 37: Maybe this variable can be `const` too. And more variables.

Is there a guideline about the usage of `const` in samples?

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/GODictionaryRenderer/GODictionaryRenderer.cpp
Gela: Yeah, unfortunately sample viewer code advises against using 'auto'. Personally I am a big fan of it myself!

ldanzinger: The reason for that is generally more geared towards user facing code - so that a snippet that showcases our API usage is more self explanatory. WIth this being sample viewer internals, I'm fine with using auto

Gela: @ldanzinger This is from a sample, not the sample viewer so I'm inclined to keep it as is.

tanneryould (approved):
LGTM

PR #1462, opened 2022-09-01
WMTS Layer sample data fix

Gela:
# Description

Update existing sample to use the new service url (last one is no longer live).

## Type of change

- [x] Bug fix
- [ ] Other enhancement


tanneryould (commented):
I can't get the map to load. It's just gray for whatever reason. This service may be down as well? Regardless, We'll need to update the screenshot for this sample I believe.

PR #1468, opened 2022-09-08
Update DisplayWfsLayer.pro sdk version

Gela:
# Description
Let's just say we should NOT be on 100.5 

## Type of change

- [x] Bug fix
- [ ] Other enhancement


tanneryould (approved):
Good catch!

PR #1471, opened 2022-09-09
Fix null width warnings

Gela:
# Description

This PR primarily removes the code to get the width of a null property which was causing warnings in Qt 6. I also made a few modifications to address other type errors to these samples. 

In one instance there was an error pertaining to using an implicit width, which declaring the width explicitly resolved.

## Type of change

- [x] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [x] Windows
- [ ] Android
- [ ] Linux
- [ ] macOS
- [ ] iOS

## Checklist

- [ ] Self-review of changes

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/DisplayGrid/DisplayGrid.qml
+++ b/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/DisplayGrid/DisplayGrid.qml
@@ -89,7 +89,7 @@ DisplayGridSample {
 
         MouseArea {
             anchors.fill: parent
-            onClicked: mouse.accepted = true;
+            onClicked: (mouse) => mouse.accepted = true;
         }
     }
 
@@ -129,7 +129,7 @@ DisplayGridSample {
             ComboBox {
                 id: gridTypeComboBox
                 property int modelWidth: 0
-                Layout.minimumWidth: modelWidth + leftPadding + rightPadding + indicator.width
+                Layout.minimumWidth: modelWidth + leftPadding + rightPadding
                 Layout.rightMargin: 10
                 Layout.fillWidth: true
                 model: [latlonGrid, mgrsGrid, utmGrid, usngGrid]
@@ -181,7 +181,7 @@ DisplayGridSample {
             ComboBox {
                 id: colorCombo
                 property int modelWidth: 0
-                Layout.minimumWidth: modelWidth + leftPadding + rightPadding + indicator.width
+                Layout.minimumWidth: modelWidth + leftPadding + rightPadding
                 Layout.rightMargin: 10
                 Layout.fillWidth: true
                 model: ["red", "white", "blue"]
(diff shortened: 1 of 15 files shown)


ldanzinger (commented):
Does this impact the rendering when you remove the indicator width? If you change the code to say `control.indicator.width`, does it work then? It should be a valid property on the combo box https://doc.qt.io/qt-6/qtquickcontrols2-customize.html#customizing-combobox

> [ComboBox](https://doc.qt.io/qt-6/qml-qtquick-controls2-combobox.html) consists of [background](https://doc.qt.io/qt-6/qml-qtquick-controls2-control.html#background-prop), [content item](https://doc.qt.io/qt-6/qml-qtquick-controls2-control.html#contentItem-prop), [popup](https://doc.qt.io/qt-6/qml-qtquick-controls2-combobox.html#popup-prop), [indicator](https://doc.qt.io/qt-6/qml-qtquick-controls2-combobox.html#indicator-prop), and [delegate](https://doc.qt.io/qt-6/qml-qtquick-controls2-combobox.html#delegate-prop).

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/DisplayGrid/DisplayGrid.qml
```diff
@@ -206,7 +206,7 @@ DisplayGridSample {
             ComboBox {
                 id: colorCombo2
                 property int modelWidth: 0
-                Layout.minimumWidth: modelWidth + leftPadding + rightPadding + indicator.width
+                Layout.minimumWidth: modelWidth + leftPadding + rightPadding
```
JamesMBallard, line 209:
```suggestion
                Layout.minimumWidth: modelWidth + leftPadding + rightPadding + (indicator ? indicator.width : 0)
```

I am not sure about the warnings, so this may not make sense.

Gela:
@JamesMBallard @ldanzinger It does not impact rendering at all, in all cases the value was null in the first place so removing that bit really just cleans up the console without changing the UI

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/DisplayGrid/DisplayGrid.qml
Gela: I feel like this is going to overcomplicate the code without contributing given that indicator is always null anyway

JamesMBallard:
>in all cases the value was null in the first place so removing that bit really just cleans up the console without changing the UI

Did that object even become non-null? Presumably that property binding will be reevaluated multiple times.

Also I see now my comments didn't come through with my suggestion. I meant to ask if the ternary would work to avoid the warnings initially, but be there for binding reevaluations once that object is valid. I would only want to remove all the `+ indicator.width` if we are confident it never worked throughout the entire app's existence.

ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/DisplayGrid/DisplayGrid.qml
JamesMBallard: That makes sense. Now that you've demo'd it and we can see it is always null, removing is right.

JamesMBallard (approved):

PR #1479, opened 2022-09-21
System reachability

tanneryould:
# Description

Update `System.isOnline` to `System.reachability === System.ReachabilityOnline`

## Type of change

- [x] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS


PaulAllanSturm:
I verified that this change fixes the original problem I ran into. I reproduced the original problem with the latest v.next, then switched over to tanner/system_reachability and the problem was resolved (no more "you must be online" message).

PaulAllanSturm (approved):
I'll never get used to `!==`

PR #1499, opened 2022-10-04
hook up the BusyIndicator to the mapDrawing status

PaulAllanSturm:
# Description

For the DisplayDrawingStatus sample, the BusyIndicator no longer appears on Windows. The fix is to hook up BusyIndicator's "running" state to the "mapDrawing" boolean. This gets it going when necessary.

I'm not sure when this regressed; my guess is it's due to the change from Qt 5 to Qt 6.

## Type of change

- [x] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [x] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

## Checklist

- [ ] Self-review of changes

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Maps/DisplayDrawingStatus/DisplayDrawingStatus.qml
+++ b/ArcGISRuntimeSDKQt_CppSamples/Maps/DisplayDrawingStatus/DisplayDrawingStatus.qml
@@ -71,7 +71,7 @@ DisplayDrawingStatusSample {
                 BusyIndicator {
                     anchors.horizontalCenter: parent.horizontalCenter
                     height: 60
-                    running: true
+                    running: displayDrawingStatusSample.mapDrawing
                 }
 
                 Text {


tanneryould:
This is weird and I'd like help understanding. I'm not sure why the BusyIndicator's visibility needs to be changed. Its visibility should be contingent on the visibility of the mapDrawingWindow which is controlled by the same mapDrawing boolean on line 43. If the BusyIndicator is just running constantly, but is hidden/visible because of the parent rectangle, I don't understand why this works.

Furthermore, the QML sample relies on the same logic. Is QML affected on Windows as well?

Lastly, I verified that the changes work on Android, but then again the originals worked on Android as well.

PaulAllanSturm:
This change doesn't impact the BusyIndicator's `visibility` setting at all -- that continues to operate as it always has, which is that it's totally dependent on the parent's visibility. This change  just impacts the `running` setting. Of course, a BusyIndicator that isn't running doesn't draw anything, so in that sense it looks the same as if `visibility` is false. The difference now is that when `mapDrawing` changes, the BusyIndicator both becomes visible (because of its parent), as well as it starts running. So apparently that's enough to get it going. Whereas before, the BusyIndicator was running non-stop, and simply changed visibility back and forth.

I can't say for sure, but there's probably some efficiency code in BusyIndicator to keep it from churning CPU cycles when it's running-but-invisible, and that code is what we were getting tripped up on. Now that it toggles `running` as well as going visible, everybody's happy.

tanneryould:
Cool, thanks for the explanation. Could you make this same change to the QML sample? Then I'll approve it, but I think we should wait for Gela's verification on Windows before merging.

PaulAllanSturm:
The QML version is implemented differently and doesn't have the same problem. It has an `onDrawStatusChanged` that checks `drawStatus === Enums.DrawStatusInProgress`. I guess the difference is because we're able to make a `mapDrawing` property in the C++ code that we otherwise can't do in pure QML.

tanneryould:
Hm, odd that the QML version works despite [`running: true`](https://github.com/Esri/arcgis-runtime-samples-qt/blob/v.next/ArcGISRuntimeSDKQt_QMLSamples/Maps/DisplayDrawingStatus/DisplayDrawingStatus.qml#L102) which caused it to fail in C++. Oh well, so long as it works.

PR #1523, opened 2022-10-25
Allow reachability unknown

tanneryould:
# Description

Adds a check to the network required view that allows for unknown network reachability. This workaround should fix the network required loop issue in the daily builds and further our debugging efforts.


ldanzinger (approved):
I'm ok with this. Sounds like from the doc, it's not exactly guaranteed:

https://doc-snapshots.qt.io/qt6-dev/qnetworkinformation.html#Reachability-enum

> If this value is returned then we may be connected but the OS has still not confirmed full connectivity, or this feature is not supported.

PR #1542, opened 2022-11-07
QML invalid chromacities fix

tanneryould:
Same change as https://github.com/Esri/arcgis-runtime-samples-qt/pull/1540 but for QML


PaulAllanSturm (approved):
LGTM

PR #1544, opened 2022-11-08
Add the necessary header so Local Server can be detected.

JamesMBallard:
# Description

Fix the local server detection logic.

## Type of change

- [x] Bug fix

### DIFF
--- a/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_CppSamples/CppSampleManager.cpp
+++ b/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_CppSamples/CppSampleManager.cpp
@@ -17,6 +17,7 @@
 
 #include "CppSampleManager.h"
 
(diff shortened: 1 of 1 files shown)


tanneryould (approved):
LGTM

ldanzinger:
I guess we need to add this include into https://github.com/Esri/arcgis-runtime-samples-qt/blob/v.next/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/SampleManager.cpp and https://github.com/Esri/arcgis-runtime-samples-qt/blob/v.next/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/mainSample.cpp since they use LOCALSERVER_SUPPORTED?

JamesMBallard:
Right you are!
https://github.com/Esri/arcgis-runtime-samples-qt/pull/1547/files

JamesMBallard:
I logged a separate issue to consider refactoring how these headers work so they can be included anywhere without errors. That would prevent problems like this in the future.

PR #1552, opened 2022-11-16
Portal item collision

tanneryould:
# Description

Fixes how we get the duplicate PortalItem's id by searching by item title rather than just getting the first item in the list. Previously we were just retrieving and deleting whatever item was first returned from the AGOL account (I hope we didn't delete anything important while testing...), so we would still encounter collisions on subsequent runs.

## Type of change

- [x] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/AddItemsToPortal/AddItemsToPortal.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/AddItemsToPortal/AddItemsToPortal.cpp
@@ -195,8 +195,15 @@ void AddItemsToPortal::connectUserSignals()
 
     if (m_user->items()->size() > 0)
     {
-      m_item->setItemId(m_user->items()->at(0)->itemId());
-      m_item->load();
+      for (PortalItem* item : *m_user->items())
+      {
+        if (item->title() == "Add Items Sample")
+        {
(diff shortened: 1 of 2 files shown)


ArcGISRuntimeSDKQt_CppSamples/CloudAndPortal/AddItemsToPortal/AddItemsToPortal.cpp
```diff
@@ -195,8 +195,15 @@ void AddItemsToPortal::connectUserSignals()
 
     if (m_user->items()->size() > 0)
```
PaulAllanSturm, line 196: I think the "size > 0" check is now redundant and can be removed -- the for loop will handle the "size=0" case by doing nothing.

tanneryould: Good call, thanks!

tanneryould:
We could add a case where we get a 409 error but the item title isn't found, but would that be overkill?

PaulAllanSturm:
I dunno, I think it's fine without additional 409 handling.

PaulAllanSturm (approved):

PR #1553, opened 2022-11-24
Update samples that use OfflineMapTask

sbiswas96:
# Description

A couple of the samples dealing with offline map tasks do the following:

> - Create a PortalItem m_portalItem
> - initialize m_map using m_portalItem
> - initialize m_offlineMapTask using m_portalItem

This results in the OfflineMapTask creating its own map from m_portalItem. So there ends up being 2 instances of a map. Although the samples work correctly, if users copy and paste the code, they may run into problems trying to do something outside what the sample indicates. 

## Type of change

- [x] Bug fix
- [ ] Other enhancement


lsmallwood (approved):
LGTM @sbiswas96 👍

tanneryould (approved):

tanneryould:
We can remove `m_portalItem` from the header file because it isn't needed anymore and just do the following to create a new map

```cpp
  // Create a map from the Portal Item
  m_map = new Map(new PortalItem(portal, webMapId(), this), this);
```

tanneryould (approved):

PR #1555, opened 2022-11-30
Samples .PRO fixes and warning fixes

tanneryould:
# Description

I'm honestly not sure how these were missed.

- Adds missing Info.plist files that prevented deployment to iOS from Qt.
- Adds missing deployment.pri files to two QML samples
- Updates QApplication to QGuiApplication

The Info.plist files were automatically generated from Qt and I confirmed that they work. Deployment to iOS was failing because some samples didn't have Info.plist and Qt wasn't automatically generating one. I had *.plist in my .gitignore_global so I don't think it was pushing up that file when merging new samples in. I think Qt 5 would automatically generate an Info.plist file/xcode will do the same, so the issue went undetected.

I'm not sure how deployment.pri was missing but it not being there didn't prevent building on macOS

I'm really not sure how there were still cases of QApplication.

## Type of change

- [x] Bug fix

**Platforms tested on**:

- [ ] Windows
- [x] Android
- [ ] Linux
- [x] macOS
- [x] iOS


ArcGISRuntimeSDKQt_QMLSamples/Search/FindAddress/main.cpp
```diff
@@ -38,7 +38,7 @@ int main(int argc, char *argv[])
   // 2. API key: A permanent key that gives your application access to Esri
   //    location services. Visit your ArcGIS Developers Dashboard create a new
   //    API keys or access an existing API key.
-  const QString apiKey = QString("");
+  const QString apiKey = QString("AAPK4f6b2b1d82e54f09b7462071964a2df9yVeNc6qmggJvnuViEN2mYBB0B0cg3clhnV0axP-MPub8RqhsLHAyob2Ll626YhTJ");
```
ldanzinger, line 41:
```suggestion
  const QString apiKey = QString("");
```

ldanzinger (changes requested):

ldanzinger (approved):

PR #1583, opened 2023-02-16
Remove no predictive text enum

tanneryould:
# Description

For some reason, this enum prevented our automation software from typing "c" or "k" in the TextField. I built an ipa with these changes and this solved the problem. Removing this enum does not seem to affect the functionality of the SearchBar.

## Type of change

- [x] Bug fix

**Platforms tested on**:

- [x] iOS

### DIFF
--- a/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/SearchBar.qml
+++ b/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/SearchBar.qml
@@ -19,7 +19,6 @@ import QtQuick.Controls
 TextField {
     id: searchBar
     rightPadding: height
-    inputMethodHints: Qt.ImhNoPredictiveText
     selectByMouse: true
 
     Image {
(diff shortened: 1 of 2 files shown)


ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/SearchBar.qml
```diff
@@ -19,7 +19,6 @@ import QtQuick.Controls
 TextField {
     id: searchBar
     rightPadding: height
-    inputMethodHints: Qt.ImhNoPredictiveText
```
JamesMBallard, line 22: This seems desirable to keep. We don't want the search field to attempt to auto-fill information. Also, if it starts doing that it could throw a wrench into any existing automation tests that don't expect it.

Are there any indications about why this is causing the problem?

tanneryould: The default value for [inputMethodHints](https://doc.qt.io/qt-6.2/qml-qtquick-textinput.html#inputMethodHints-prop) is `Qt.ImhNone` so I don't think it will attempt to auto-fill any information. In my tests with that enum removed it didn't show any predictive text.
As for why this is causing the problem, I do not know. I had Sarat test a different sample (Feature Layer Query) that has a TextField box with this enum and that works fine so this enum is probably not directly to blame. I'll do some more digging on this before committing anything.

ldanzinger (approved):
LGTM

PR #1584, opened 2023-02-22
Update AR samples for Qt 6

tanneryould:
# Description

Updates AR samples to Qt 6. They compile and run, but are unusably slow which will require a follow up.

**Platforms tested on**:

- [x] Android
- [x] iOS


ldanzinger (approved):
LGTM

PR #1592, opened 2023-04-04
Further enum fixes

tanneryould:
# Description

I caught a few more cases where strings are passed to properties that require enums


ldanzinger (approved):
good catch, thanks

PR #1595, opened 2023-04-13
Tanneryould/sv downloader

tanneryould:
# Description

This PR adds the following:

- A cancel button to the "Offline Data Manager View" that allows the user to cancel any remaining downloads in the queue. Previously, if the user clicked "download all offline data" it would begin the download and the only way to cancel would be to close the application. Even if you close the data download manager view, the next sample that would require offline data would just show the offline download in progress.
- If a download fails, it will display the item id that failed, and a retry button. Previously a failed download would also prevent the user from downloading other offline data afterwards because the app state would get stuck on that download failed page. Now it'll be able to download other offline data even after a failed download.
- It also moves the line to propagate the description view to before the check to see if all offline data for a sample is present, so a user can see a sample's description even if they haven't downloaded the offline data.

## Type of change

- [ ] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

### DIFF
--- a/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/SampleManager.h
+++ b/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/SampleManager.h
@@ -52,6 +52,8 @@ class SampleManager : public QObject
   Q_PROPERTY(bool downloadInProgress READ downloadInProgress WRITE setDownloadInProgress NOTIFY downloadInProgressChanged)
   Q_PROPERTY(QString downloadText READ downloadText WRITE setDownloadText NOTIFY downloadTextChanged)
   Q_PROPERTY(double downloadProgress READ downloadProgress WRITE setDownloadProgress NOTIFY downloadProgressChanged)
+  Q_PROPERTY(bool cancelDownload READ cancelDownload WRITE setCancelDownload NOTIFY cancelDownloadChanged)
+  Q_PROPERTY(bool downloadFailed READ downloadFailed WRITE setDownloadFailed NOTIFY downloadFailedChanged)
 
 public:
   explicit SampleManager(QObject* parent = nullptr);
@@ -77,11 +79,13 @@ public:
 
 signals:
   void sampleInitComplete();
+  void cancelDownloadChanged();
   void currentModeChanged();
   void currentSampleChanged();
   void currentCategoryChanged();
   void currentSourceCodeChanged();
   void doneDownloadingChanged();
+  void downloadFailedChanged();
   void downloadInProgressChanged();
   void downloadTextChanged();
   void downloadProgressChanged();
@@ -117,6 +121,10 @@ private:
   void setDownloadText(const QString& downloadText);
   double downloadProgress() const { return m_downloadProgress; }
   void createAndSetTempDirForLocalServer();
+  bool cancelDownload() const { return m_cancelDownload; }
+  void setCancelDownload(bool cancel);
+  bool downloadFailed() const { return m_downloadFailed; }
+  void setDownloadFailed(bool didFail);
 
 private:
   CategoryListModel* m_categories = nullptr;
@@ -132,6 +140,8 @@ private:
   QString m_downloadText = QString("Downloading");
   double m_downloadProgress = 0.0;
   std::unique_ptr<QTemporaryDir> m_tempDir;
+  bool m_cancelDownload = false;
+  bool m_downloadFailed = false;
 };
 
 #endif // SAMPLEMANAGER_H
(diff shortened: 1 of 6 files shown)


ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/SampleManager.h
```diff
@@ -52,6 +52,8 @@ class SampleManager : public QObject
   Q_PROPERTY(bool downloadInProgress READ downloadInProgress WRITE setDownloadInProgress NOTIFY downloadInProgressChanged)
   Q_PROPERTY(QString downloadText READ downloadText WRITE setDownloadText NOTIFY downloadTextChanged)
   Q_PROPERTY(double downloadProgress READ downloadProgress WRITE setDownloadProgress NOTIFY downloadProgressChanged)
+  Q_PROPERTY(bool cancelDownload READ cancelDownload WRITE setCancelDownload NOTIFY cancelDownloadChanged)
```
ldanzinger, line 55: should it be `downloadCanceled`? `cancelDownload` sounds more like an action or method

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/DataDownloadView.qml
```diff
@@ -48,7 +48,13 @@ Page {
...
 
         Label {
-            text: qsTr("This sample uses offline data that is not detected on your system. Press the 'Download' button to download the data to your device.")
+            text: {
+                if (SampleManager.downloadFailed) {
+                    SampleManager.downloadText;
+                } else {
+                    "This sample uses offline data that is not detected on your system. Press the 'Download' button to download the data to your device.";
+                }
+            }
```
ldanzinger, line 57: a ternary operator would be preferable 

```suggestion
            text:  SampleManager.downloadFailed ? SampleManager.downloadText : "This sample uses offline data that is not detected on your system. Press the 'Download' button to download the data to your device."
```

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/ManageOfflineDataView.qml
```diff
@@ -118,6 +124,19 @@ Page {
...
         }
+
+        Button {
+            // PortalItem::fetchData does not have a cancel method so we can only clear the remaining items from the download queue
+            text: SampleManager.cancelDownload ? "Cancelling remaining downloads..." : "Cancel remaining downloads"
+            Layout.alignment: Qt.AlignHCenter | Qt.AlignVCenter
+            visible: SampleManager.downloadInProgress
+            enabled: !SampleManager.cancelDownload
+            onClicked: {
+                console.log("attempting to cancel download");
```
ldanzinger, line 135: you've got a couple of console.logs in here, perhaps want to clean them up unless they are important to keep

ldanzinger (approved):
LGTM. Got a couple of cleanup suggestions to consider

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/SampleManager.h
tanneryould: When downloading all offline data, after each download finishes, we check to see if `cancelDownload` is true or not before proceeding to download the next item in the downloadQueue. I opted to word is that way because it's like a flag to take an action in the future, whereas `downloadCanceled` implies to me that something happened already.

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/main.qml
```diff
@@ -318,18 +294,26 @@ ApplicationWindow {
         }
 
         function onDoneDownloadingChanged() {
-            SampleManager.currentMode = SampleManager.LiveSampleView;
             showSample();
         }
+
+        function onCurrentModeChanged() {
+            console.log("current mode changed:",
+                        ["LiveSampleView","SourceCodeView","DescriptionView","ManageOfflineDataView","NetworkRequiredView","DownloadDataView"][SampleManager.currentMode])
```
tanneryould, line 302:
```suggestion
```

ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/ManageOfflineDataView.qml
```diff
@@ -118,6 +124,19 @@ Page {
...
+                SampleManager.cancelDownload = true;
```
tanneryould, line 136:
```suggestion
                SampleManager.cancelDownload = true;
```

tanneryould (commented):
Prep to apply suggestions from code review

PR #1607, opened 2023-05-23
Qt 6.5 UI Fixes

tanneryould:
# Description

Qt 6.5 introduces changes to the Material theme that our Sample Viewers use. Some of these changes break our UIs a bit, so this PR updates affected UIs to work with this new theme. Most commonly affected were the ComboBox components. I added a background rectangle where necessary and also added some width padding to the boxes to account for the dropdown arrows. I'm opening this PR after testing on macOS but will test on all platforms.

## Type of change

- [x] Bug fix

**Platforms tested on**:

- [x] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/DisplayGrid/DisplayGrid.qml
+++ b/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/DisplayGrid/DisplayGrid.qml
@@ -129,7 +129,7 @@ DisplayGridSample {
             ComboBox {
                 id: gridTypeComboBox
                 property int modelWidth: 0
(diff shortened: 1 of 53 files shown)


ldanzinger (commented):
Looking good @tanneryould. Couple questions:

* Instead of picking 20 as the number, can we get the width of the combo box's indicator? That might be more future proof to changes down the line.
* Also, could I see a screenshot inside and outside of the sample viewer?

tanneryould:
@ldanzinger Here's a comparison of the Spatial Operations sample as a standalone sample on the left and in the sample viewer on the right.
Screenshot 2023-05-24 at 12 23 46 PM

ldanzinger (approved):

PR #1608, opened 2023-05-24
Don't initialize sample viewer with a sample

tanneryould:
# Description

This PR makes it so the sample viewer no longer starts with an initial sample loaded in the background. Therefore the live sample view, description view, and source code view, all start disabled as well.

## Type of change

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

### DIFF
--- a/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/SampleListView.qml
+++ b/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_Samples/qml/SampleListView.qml
@@ -68,7 +68,7 @@ Page {
         ListView {
             anchors.fill: parent
             clip: true
-            model: SampleManager.currentCategory.samples
+            model: SampleManager.currentCategory ? SampleManager.currentCategory.samples : []
             spacing: 10
             delegate: ItemDelegate {
                 id: itemDelegate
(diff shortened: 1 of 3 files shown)


tanneryould:
Should we have these disabled or hidden? I'm not sure. Will removing them affect automation in any way?
Screenshot 2023-05-24 at 2 10 57 PM

ldanzinger (approved):
LGTM. Please confirm if you've tested all the different scenarios. For example, this change also makes it so no category is selected - if you no longer have a category selected by default, and then straight away choose a featured sample or use search but never select a category, does anything blow up?

ldanzinger:
@tanneryould good question. Hidden would probably be better. I'm not sure if this would affect automation. I assume not, since when a sample is opened, it should behave as it did before. I doubt the automation would try to check description immediately upon opening, but could be wrong. Please check with Harini and Kathleen

tanneryould:
> LGTM. Please confirm if you've tested all the different scenarios. For example, this change also makes it so no category is selected - if you no longer have a category selected by default, and then straight away choose a featured sample or use search but never select a category, does anything blow up?

I've tested the following in both C++ and QML, and no issues seem to arise

- Open a featured sample
- Open a sample from search
- Open a sample from a category
- Open a sample requiring authentication
- Open a sample that requires offline data
- Download all offline data and interrupt the downloads
- Download all offline data to completion

> good question. Hidden would probably be better. I'm not sure if this would affect automation. I assume not, since when a sample is opened, it should behave as it did before. I doubt the automation would try to check description immediately upon opening, but could be wrong. Please check with Harini and Kathleen

According to Kathleen, "automation doesn't check for description upon opening, so it should be fine."

tanneryould:
@ldanzinger I had to also null the height of the menu items when hiding them, otherwise we get the result in the first picture. The latter two pictures are how the sample viewer behaves after this most recent commit.
Screenshot 2023-05-24 at 2 44 49 PM
Screenshot 2023-05-24 at 2 45 10 PM
Screenshot 2023-05-24 at 2 45 17 PM

ldanzinger (approved):
Nice work

PR #1611, opened 2023-05-30
Update WMS layer URL

tanneryould:
# Description

The WMS Url Layer service was discontinued so we needed to update it to this new URL.

### DIFF
--- a/ArcGISRuntimeSDKQt_QMLSamples/Layers/WmsLayerUrl/WmsLayerUrl.qml
+++ b/ArcGISRuntimeSDKQt_QMLSamples/Layers/WmsLayerUrl/WmsLayerUrl.qml
@@ -39,8 +39,8 @@ Rectangle {
 
             // Add a WMS Layer by specifying the URL and layer name
             WmsLayer {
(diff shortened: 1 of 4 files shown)


ArcGISRuntimeSDKQt_QMLSamples/Layers/WmsLayerUrl/WmsLayerUrl.qml
```diff
@@ -39,8 +39,8 @@ Rectangle {
 
             // Add a WMS Layer by specifying the URL and layer name
             WmsLayer {
-                url: "https://nowcoast.noaa.gov/arcgis/services/nowcoast/radar_meteo_imagery_nexrad_time/MapServer/WMSServer?request=GetCapabilities&service=WMS"
-                layerNames: ["1"]
```
ldanzinger, line 43: wonder why the layer names differed between c++ and qml

ldanzinger (approved):
LGTM

ArcGISRuntimeSDKQt_QMLSamples/Layers/WmsLayerUrl/WmsLayerUrl.qml
tanneryould: I'm not sure I understand, it looks like they were both "1" before and now they're both "conus_base_reflectivity_mosaic"?

PR #1614, opened 2023-06-12
Use unit factory in geodesic buffer and fix enum typo

tanneryould:
# Description

The Geodetic buffer was not displaying because we were passing in an enum for the linear units rather than the QML Object that the function expects. Is this an expected change?

Previous error message:
```
"Could not convert argument 2 at"
	 "bufferPoint@qrc:/Samples/Geometry/Buffer/Buffer.qml:123"
	 "@qrc:/Samples/Geometry/Buffer/Buffer.qml:47"
qrc:/Samples/Geometry/Buffer/Buffer.qml:123: TypeError: Passing incompatible arguments to C++ functions from JavaScript is not allowed.
```

## Type of change

- [x] Bug fix

**Platforms tested on**:

- [x] macOS


ldanzinger (approved):
LGTM

PR #1619, opened 2023-06-27
No need to enforce OpenGL anymore.

JamesMBallard:
# Description

Remove the enforcement of OpenGL in the daily sample builds. This will light up Direct3D on Windows and Metal for macOS/iOS.

Same change applies to individual samples.

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/AR/DisplayScenesInTabletopAR/main.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/AR/DisplayScenesInTabletopAR/main.cpp
@@ -29,9 +29,6 @@
 
 int main(int argc, char *argv[])
 {
-  // Enforce OpenGL
-  qputenv("QSG_RHI_BACKEND", "opengl");
-
   // There are some conflicts between the AR frameworks and Qt's rendering thread.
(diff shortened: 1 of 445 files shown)


GuillaumeBelz (approved):
A lot of samples! :)

LGFM

ldanzinger (approved):
spot checked, LGTM

tanneryould (approved):
LGTM, though do we need to remove OpenGL from the .pro files as well? For example: https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/v.next/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/DisplayGrid/DisplayGrid.pro#L25 has 

```
# additional modules are pulled in via arcgisruntime.pri
QT += opengl qml quick
```

JamesMBallard:
@tanneryould that's a good question. I will log a new issue to look into that, but my hunch says that we still need it. All samples still retain the ability to support OpenGL even if they don't by default.

GuillaumeBelz:
> we still need it

Maybe not. `opengl` module is used for `QOpenGLxxx` features, which is used but internally. I guess this is not used in samples and may be removed.

PR #1234, opened 2021-06-10
Fixed mapView->locationDisplay() crash involving QTimers

MicahSamaduroff:
Reloading samples involving the `locationDisplay()` caused a crash on the reload, from stopping QTimers from a different thread than they were constructed on. 

The underlying issue is being worked on, and this is a temporary fix to prevent the crash until it is complete.

PR #1630, opened 2023-08-04
Disable the location display when switching between samples.

JamesMBallard:
Workaround for an internal timer problem.

# Description

Same workaround as https://github.com/Esri/arcgis-maps-sdk-samples-qt/pull/1234. I tested locally on iOS with the QML sample app viewer and it no longer crashes when switching away from the indoors sample.

## Type of change

- [X] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [X] iOS


tanneryould (approved):
Looks good, I did a similar thing with other device positioning samples.

PR #1632, opened 2023-08-07
The MapView is not declared within the QML sample.

JamesMBallard:
We need to make sure the destruction order is like we expect.

# Description

This sample is different than the others that use this workaround. In the others, the QML is declared like this:
```qml
MyCoolSample {
    MapView { ...
```

But this one is:
```qml
Item {
    MapView { ... }

    ShowDeviceLocationUsingIndoorPositioningSample {
        id: model
        mapView: view
    }
}
```

So the sample's lifetime is not tied to the MapView's lifetime. It was crashing because the MapView was partially destructed when we get to the sample's destructor.

I tested this new fix on both macOS and iOS.

## Type of change

- [X] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [X] macOS
- [X] iOS


tanneryould (approved):
Good catch!

PR #1640, opened 2023-08-21
Add scene filtering to v.next

tanneryould:
Adds the scene filtering sample to v.next and updates it to 200.3.0.

@JamesMBallard it looks like some template version numbers were missed that I picked up in updating to 200.3.0. Want me to include these or would you want a separate PR?


JamesMBallard:
@tanneryould please include them here. Sorry about that.

bri12415 (approved):

PR #1653, opened 2023-09-18
James/port samples edit data features

JamesMBallard:
# Description

This is part 1 of 2. The file list is getting long so I wanted to break it up and make it easier to review.

This ports all the async code in all "Edit Data" category of samples to the new async pattern.


JamesMBallard:
I have a github action check that failed, but I didn't make any changes to that file that should impact or cause that alert to appear. I tried to fix it, but it still fails. I'm going to merge anyway with the error. I suspect that file would fail the check even I didn't make any changes to it.

PR #1665, opened 2023-11-14
fix path to cpp project file

GuillaumeBelz:
Some fix. `toolkitcpp.pri` is included but `toolkitqml.pri` is checked. I guess it's just a wrong copy-paste.


JamesMBallard (approved):
Good catch

PR #1668, opened 2023-11-28
Fix crash.

JamesMBallard:
I shuffled around some existing code to be sure we handle the "no results" case visually as well, by clearing out any existing callout and graphic pin that was already placed from a previous geocode.

# Description

## Type of change

- [x] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [x] macOS

### DIFF
--- a/ArcGISRuntimeSDKQt_CppSamples/Search/OfflineGeocode/OfflineGeocode.cpp
+++ b/ArcGISRuntimeSDKQt_CppSamples/Search/OfflineGeocode/OfflineGeocode.cpp
@@ -319,7 +319,7 @@ void OfflineGeocode::onGeocodingCompleted_(const QList<GeocodeResult>& geocodeRe
   m_pinGraphic->setVisible(false);
 
   // if there are no matching results, notify user and stop processing
-  if (geocodeResults.empty())
+  if (geocodeResults.isEmpty())
   {
     m_noResults = true;
     emit noResultsChanged();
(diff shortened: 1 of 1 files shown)


ArcGISRuntimeSDKQt_CppSamples/Search/OfflineGeocode/OfflineGeocode.cpp
```diff
@@ -314,19 +314,22 @@ void OfflineGeocode::onGeocodingCompleted_(const QList<GeocodeResult>& geocodeRe
...
   emit geocodeInProgressChanged();
 
-  // if there are no matching results, notify user
-  if (geocodeResults.empty())
+  // dismiss callouts
+  m_calloutData->setVisible(false);
+  m_pinGraphic->setVisible(false);
+
+  // if there are no matching results, notify user and stop processing
+  if (geocodeResults.isEmpty())
```
JamesMBallard, line 322: I changed to `isEmpty` purely for aesthetics. It is the same API and played no part in the problems.

https://doc.qt.io/qt-6/qlist.html#empty

ArcGISRuntimeSDKQt_CppSamples/Search/OfflineGeocode/OfflineGeocode.cpp
```diff
@@ -314,19 +314,22 @@ void OfflineGeocode::onGeocodingCompleted_(const QList<GeocodeResult>& geocodeRe
...
   {
     m_noResults = true;
     emit noResultsChanged();
+    return;
```
JamesMBallard, line 326: This is the real fix. Obviously we should call `geocodeResults.at(0)` after we've determined the QList is empty.

PR #1722, opened 2024-06-18
Update for LoadableImage disk path

aji13187:
# Description

Change is made to update the local url of the image (rather than the portal url) when the image is actually loaded and saved to disk.
The `BasemapStyleInfo::thumbnailUrlChanged()` is emitted when the local path of the image is available.

## Type of change

- [ ] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

## Checklist

- [x] Self-review of changes
- [x] There is no leftover commented code
- [x] Screenshots are correct size and display in description tab (500px by 500px, platform agnostic)

### DIFF
--- a/CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
+++ b/CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
@@ -59,13 +59,13 @@ void BasemapStyleListModel::insertItemsIntoGallery(const QList<BasemapStyleInfo*
 {
     beginResetModel();
     m_previews.clear();
-    for (const BasemapStyleInfo* info : infos)
+    for (auto index = 0; index < infos.size(); ++index)
     {
-        connect(info, &BasemapStyleInfo::thumbnailUrlChanged, this, [this, info]() {
-            QModelIndex index = createIndex(m_previews.indexOf(info), 0);
-            emit dataChanged(index, index);
+        m_previews.insert(index, std::move(infos.at(index)));
+        QModelIndex modelIndex = createIndex(index, 0);
+        connect(m_previews.at(index), &BasemapStyleInfo::thumbnailUrlChanged, this, [this, modelIndex]() {
+            emit dataChanged(modelIndex, modelIndex);
         });
-        m_previews.append(info);
     }
     endResetModel();
 }
--- a/CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
+++ b/CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
@@ -62,8 +62,9 @@ void BasemapStyleListModel::insertItemsIntoGallery(const QList<BasemapStyleInfo*
     for (auto index = 0; index < infos.size(); ++index)
     {
         m_previews.insert(index, infos.at(index));
-        QModelIndex modelIndex = createIndex(index, 0);
-        connect(m_previews.at(index), &BasemapStyleInfo::thumbnailUrlChanged, this, [this, modelIndex]() {
+        connect(m_previews.at(index), &BasemapStyleInfo::thumbnailUrlChanged, this, [this, index]() 
+        {
+            const QModelIndex modelIndex = createIndex(index, 0);
             emit dataChanged(modelIndex, modelIndex);
         });
     }
--- a/CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
+++ b/CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
@@ -64,7 +64,7 @@ void BasemapStyleListModel::insertItemsIntoGallery(const QList<BasemapStyleInfo*
         m_previews.insert(index, infos.at(index));
         connect(m_previews.at(index), &BasemapStyleInfo::thumbnailUrlChanged, this, [this, index]() 
         {
-            const QModelIndex modelIndex = createIndex(index, 0);
+            QModelIndex modelIndex = createIndex(index, 0);
             emit dataChanged(modelIndex, modelIndex);
         });
     }
--- a/CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
+++ b/CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
@@ -51,7 +51,7 @@ QVariant BasemapStyleListModel::data(const QModelIndex& index, int role) const
 
 int BasemapStyleListModel::rowCount(const QModelIndex& parent) const
 {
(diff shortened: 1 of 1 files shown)


aji13187:
This wont make any difference in the output of the thumbnail images on the screen.
Not sure this will confuse user or need comments.

CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
```diff
@@ -60,7 +61,11 @@ void BasemapStyleListModel::insertItemsIntoGallery(const QList<BasemapStyleInfo*
     m_previews.clear();
     for (const BasemapStyleInfo* info : infos)
     {
-        m_previews.append(std::move(info));
+        connect(info, &BasemapStyleInfo::thumbnailUrlChanged, this, [this, info]() {
```
GuillaumeBelz, line 64: Is `info` lifetime is long enough to be used in connection? Maybe moving the index outside the lambda function is safer?

```cpp
const auto index = m_previews.indexOf(info);
connect(info, &BasemapStyleInfo::thumbnailUrlChanged, this, [this, index]() {
...
```

CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
```diff
@@ -51,6 +51,7 @@ QVariant BasemapStyleListModel::data(const QModelIndex& index, int role) const
 
 int BasemapStyleListModel::rowCount(const QModelIndex& parent) const
 {
+    Q_UNUSED(parent);
```
JamesMBallard, line 54:
```suggestion
    Q_UNUSED(parent)
```

CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
JamesMBallard: I like @GuillaumeBelz's idea, but I would augment that to add that we should just switch to a traditional for loop with index instead of repeatedly calling `indexOf` which is not a cheap call.

JamesMBallard:
>This wont make any difference in the output of the thumbnail images on the screen.
>Not sure this will confuse user or need comments.

If you are confident this makes no difference in the workflow, we should not do it. We don't want to overcomplicate the sample code if we can avoid it.

JamesMBallard (commented):
One comment to add on to what Guillaume already said, but we should consider if this is worthwhile to do if it makes no difference.

CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
aji13187: Used traditional for loop

aji13187:
> > This wont make any difference in the output of the thumbnail images on the screen.
> > Not sure this will confuse user or need comments.
> 
> If you are confident this makes no difference in the workflow, we should not do it. We don't want to overcomplicate the sample code if we can avoid it.

The only difference is now is the sample is using the disk url rather than the network portal image url. This will improve the loading speed the thumbnails if more reloading of thumbnails is required in user apps, but not really improves anything in this sample since we do not reload the page.
It may help the user, so I think we can keep this change.

CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
```diff
@@ -51,16 +51,21 @@ QVariant BasemapStyleListModel::data(const QModelIndex& index, int role) const
...
 
 void BasemapStyleListModel::insertItemsIntoGallery(const QList<BasemapStyleInfo*> infos)
 {
     beginResetModel();
     m_previews.clear();
-    for (const BasemapStyleInfo* info : infos)
+    for (auto index = 0; index < infos.size(); ++index)
     {
-        m_previews.append(std::move(info));
+        m_previews.insert(index, std::move(infos.at(index)));
```
JamesMBallard, line 64:
```suggestion
        m_previews.insert(index, infos.at(index));
```
Guillaume probably knows better than I do, but I assume infos.at is already an r-value so no need to std::move it.

CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
```diff
@@ -51,16 +51,21 @@ QVariant BasemapStyleListModel::data(const QModelIndex& index, int role) const
...
+        QModelIndex modelIndex = createIndex(index, 0);
+        connect(m_previews.at(index), &BasemapStyleInfo::thumbnailUrlChanged, this, [this, modelIndex]() {
+            emit dataChanged(modelIndex, modelIndex);
+        });
```
JamesMBallard, line 68: Use consistent bracket style. Also is it possible to just capture the index initially rather than the QModelIndex?
```suggestion
        connect(m_previews.at(index), &BasemapStyleInfo::thumbnailUrlChanged, this, [this, index]() 
        {
            const QModelIndex modelIndex = createIndex(index, 0);
            emit dataChanged(modelIndex, modelIndex);
        });
```

This code may not work, proceed with caution.

CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
aji13187: Looks like QList::at() is not rvalue, but still `std::move` is not necessary because of Qt's implicit sharing or copy-on-write mechanism.

CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
aji13187: Tested and working.

CppSamples/Maps/CreateDynamicBasemapGallery/BasemapStyleListModel.cpp
GuillaumeBelz: > infos.at is already an r-value

It's tricky. An function call is a lvalue if the return type is a lvalue-reference (including const) and a prvalue otherwise. https://en.cppreference.com/w/cpp/language/value_category

`QList::at` returns a const ref (https://doc.qt.io/qt-6/qlist.html#at) so it's a lvalue. Using `std::move` change nothing. 

But there is no copy, because `QList::at` returns a const ref and `QList::insert` takes a const ref.

TL;DR: `const` can be removed

aji13187: Thanks, removed `const`.

JamesMBallard (approved):

GuillaumeBelz (approved):

PR #1733, opened 2024-07-09
Jared/api key to access token

jared-2016:
# Description

update apikey terminology to use access token instead. Also renamed samples when being run standalone to not have ` - C++` included in it's name.

Also caught a few main signatures that deviated from the rest. Updated them so it's unified.

since i modified main.cpp files of individual samples, I picked 1/2 from each category and built them standalone and ran into no issues. Testing all 220 I think is unnecessary as the changes are all the same.

## Type of change

- [ ] Bug fix
- [x] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

## Checklist

- [x] Self-review of changes

### DIFF
--- a/CppSamples/AR/DisplayScenesInTabletopAR/main.cpp
+++ b/CppSamples/AR/DisplayScenesInTabletopAR/main.cpp
@@ -51,22 +51,30 @@ int main(int argc, char *argv[])
   QGuiApplication app(argc, argv);
   app.setApplicationName(QStringLiteral("DisplayScenesInTabletopAR - C++"));
 
-  // Use of Esri location services, including basemaps and geocoding,
-  // requires authentication using either an ArcGIS identity or an API Key.
-  // 1. ArcGIS identity: An ArcGIS named user account that is a member of an
-  //    organization in ArcGIS Online or ArcGIS Enterprise.
-  // 2. API key: A permanent key that gives your application access to Esri
-  //    location services. Visit your ArcGIS Developers Dashboard create a new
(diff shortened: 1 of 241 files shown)


jared-2016:
The templates have not been included in this PR as of right now.

The templates use a helper function to set the apikey(access token) but it also checks the command line to see if it was passed in that way. If this is something we wish to keep. It will require some reformatting from what the original issue requested it to be changed too. It's doable to find and replace it and put the forward declare and function implementation back. Seems like a luxury that I don't find most people doing.

for reference: https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/7c5d595b96a23f3deef70d59401413eb4f6170aa/Templates/CppSampleTemplate/main.cpp.tmpl#L85-L106

jared-2016:
Also it's worth mentioning that it was never rolled out to all samples. We just updated the template at some point and started using it from there. If this is something we should decide to keep, we should update the samples to use this so the experience is consistent.

JamesMBallard:
I agree with your assessment to just remove the command-line method of setting the API key. It was only there for a subset of samples, so we should make it official and consistent in the future if we want it. Good call.

JamesMBallard (approved):
I did a spot check. Looks good.

jared-2016:
I updated templates which made me realize I forgot Widgets 😮

ldanzinger (approved):
LGTM

PR #1740, opened 2024-07-22
James/restore ls samples

JamesMBallard:
# Description

Add LocalServer category back to the sample viewer for Windows and Linux, if and only if local server is installed on the machine it's running on.

This logic was lost in the porting. There is no more `CppSampleManager` class, so I removed `virtual` to make it clearer there's no more overrides. 
https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/6d49d3fce7f8b89d72da7589aacd22c90d91f773/ArcGISRuntimeSDKQt_SampleViewers/ArcGISRuntimeSDKQt_CppSamples/CppSampleManager.cpp#L66-L77

## Type of change

- [x] Bug fix

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [x] Linux
- [ ] macOS
- [ ] iOS

### DIFF
--- a/SampleViewer/SampleManager.cpp
+++ b/SampleViewer/SampleManager.cpp
@@ -139,7 +139,7 @@ void SampleManager::setDownloadFailed(bool didFail)
 // Build the Categories List
 void SampleManager::buildCategoriesList()
 {
-  QDir dir(DIRNAMESAMPLES);
+  const QDir dir(DIRNAMESAMPLES);
   QFile xmlFile(":/Samples/Categories.xml");
 
   // Go through the XML to find the sample categories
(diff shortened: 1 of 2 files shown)


SampleViewer/SampleManager.cpp
```diff
@@ -148,23 +148,30 @@ void SampleManager::buildCategoriesList()
...
   {
     qWarning() << "could not find Categories.xml";
   }
+
+#ifdef LOCALSERVER_SUPPORTED
+  if (LocalServer::isInstallValid())
+  {
+    appendCategoryToManager(createCategory("LocalServer", "Local Server", dir));
+  }
+#endif // LOCALSERVER_SUPPORTED
```
JamesMBallard, line 174: This block is the fix. The rest are unrelated cleanup.

jared-2016 (approved):

PR #1745, opened 2024-08-01
James/sample include ordering

JamesMBallard:
# Description

Change include pattern. We were approaching the limits of the command line length on Windows. We have hit multiple problems over the years related to this. The compile command on macOS for DataItem.cpp goes from 28281 characters to  5250 with these changes.

We now include sample headers by relative path in mainSample.cpp (the only place they're needed), but the entire sample app viewer had been including the full include path to every individual sample, which is unsustainable.

This PR includes other cleanup, like sorting long lists of calls and includes.

ListRelatedFeatures sample had some minor refactoring to provide a property of QAbstractListModel rather than RelatedFeatureListModel. I tested it locally and it still worked fine.

## Type of change

- [X] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [X] Windows
- [x] Android
- [ ] Linux
- [X] macOS
- [x] iOS


SampleViewer/mainSample.cpp
```diff
@@ -45,252 +46,244 @@
...
-#include "ViewshedCamera.h"
-#include "ViewshedGeoElement.h"
-#include "ViewshedLocation.h"
-#include "Web_Tiled_Layer.h"
-#include "WmsLayerUrl.h"
-#include "WMTS_Layer.h"
+#include "../CppSamples/Analysis/AnalyzeHotspots/AnalyzeHotspots.h"
+#include "../CppSamples/Analysis/AnalyzeViewshed/AnalyzeViewshed.h"
+#include "../CppSamples/Analysis/DistanceMeasurementAnalysis/DistanceMeasurementAnalysis.h"
+#include "../CppSamples/Analysis/Geotriggers/Geotriggers.h"
```
JamesMBallard, line 59: This entire block (and below) are where the fix is. Since we no longer add each sample's path to the compiler include path, we need to reference each one by relative path now.

SampleViewer/mainSample.cpp
```diff
@@ -574,9 +567,9 @@ void registerCppSampleClasses()
   ViewshedCamera::init();
   ViewshedGeoElement::init();
   ViewshedLocation::init();
+  WMTS_Layer::init();
   Web_Tiled_Layer::init();
   WmsLayerUrl::init();
-  WMTS_Layer::init();
```
JamesMBallard, line 579: I sorted this entire section of `init` calls, so that why there are changes.

SampleViewer/mainSample.cpp
jared-2016: Just an idea but could we create a separate header, say `CppSamplesRelativePaths.h`, and put this block of sample header includes there?

I did see for a similar idea you mentioned that it's just one more place to remember which i agree. It just gets under my skin that we have wall of header includes for the samples and we also have a wall of `::init` for each sample. Between those two, it makes up about 350 lines of 635 lines of this file.

jared-2016 (approved):
Looks good! Just one comment/idea I wanted to throw out there.

SampleViewer/mainSample.cpp
JamesMBallard: It's a good point, but I think in this case it has to go somewhere, so we're not saving much by putting it in a separate file. We know those headers are needed in this file, but if someone is perusing the repo, it will not be clear what the purpose of `CppSamplesRelativePaths.h` is by itself. So I lean towards leaving them here.

.gitignore
```diff
@@ -141,8 +141,10 @@ Thumbs.db
 
 # ignore all build directories
 
-*build*
+*build-*
 .qmake.stash
 *Makefile*
 .qtc_clangd
 mocinclude.opt
+build/
```
JamesMBallard, line 149: FYI @jared-2016 I snuck this change in. I was getting warnings that files were being ignored that we probably didn't intend, like CppSamples/DisplayInformation/BuildLegend and CppSamples/Maps/BrowseBuildingFloors.

PR #1746, opened 2024-08-01
James/200.6.0

JamesMBallard:
Update v.next to 200.6


Templates/CppWidgetsSampleTemplate/wizard.json
JamesMBallard, line 1: Lots of changes to this file. That is because it has a mish-mash of tabs and spaces and various indents. I did the typical 200.5 -> 200.6 version changes and also auto-formatted the entire file to get rid of all the spacing inconsistencies.

jared-2016 (approved):
Quick spot check of everything under `CppSamples` and `CppWidgetsSamples` since it was just repeated update of version. Everything else I looked more closely and it all looks good!

PR #1748, opened 2024-08-12
James/task manager

JamesMBallard:
# Description

During sample automation runs, the UN trace and valve isolation trace samples are hitting problems where the sample is destructing at the same time as async operations are completing. We can proactively protect against this by tracking and cleaning up 

This is overkill to deploy to every sample, but it does provide a simple mechanism where needed to cancel running async calls.

## Type of change

- [X] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

### DIFF
--- a/CppSamples/UtilityNetwork/PerformValveIsolationTrace/PerformValveIsolationTrace.cpp
+++ b/CppSamples/UtilityNetwork/PerformValveIsolationTrace/PerformValveIsolationTrace.cpp
@@ -257,12 +257,10 @@ void PerformValveIsolationTrace::performTrace()
       traceParameters->traceConfiguration()->filter()->setBarriers(categoryComparison);
     }
 
-    m_taskCanceler->addTask(
-      m_utilityNetwork->traceAsync(traceParameters).then(this, [this](QList<UtilityTraceResult*>)
-      {
-        onTraceCompleted_();
-      })
-    );
+    m_taskCanceler->addTask(m_utilityNetwork->traceAsync(traceParameters).then(this, [this](QList<UtilityTraceResult*>)
+    {
+      onTraceCompleted_();
+    }));
   }
 }
 
--- a/CppSamples/UtilityNetwork/PerformValveIsolationTrace/PerformValveIsolationTrace.cpp
+++ b/CppSamples/UtilityNetwork/PerformValveIsolationTrace/PerformValveIsolationTrace.cpp
@@ -69,6 +69,9 @@
 #include <QFuture>
 #include <QUuid>
 
+// Other headers
+#include "TaskCanceler.h"
+
 using namespace Esri::ArcGISRuntime;
 
 namespace
@@ -103,7 +106,8 @@ PerformValveIsolationTrace::PerformValveIsolationTrace(QObject* parent /* = null
   m_startingLocationOverlay(new GraphicsOverlay(this)),
   m_filterBarriersOverlay(new GraphicsOverlay(this)),
   m_serviceGeodatabase(new ServiceGeodatabase(featureServiceUrl, m_cred, this)),
-  m_graphicParent(new QObject())
+  m_graphicParent(new QObject()),
+  m_taskCanceler(std::make_unique<TaskCanceler>())
 {
   // disable UI while loading service geodatabase and utility network
   m_tasksRunning = true;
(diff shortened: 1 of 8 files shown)


CppSamples/UtilityNetwork/TraceUtilityNetwork/TraceUtilityNetwork.cpp
```diff
@@ -186,10 +190,11 @@ void TraceUtilityNetwork::connectSignals()
     constexpr double tolerance = 10.0;
     constexpr bool returnPopups = false;
     m_clickPoint = m_mapView->screenToLocation(mouseEvent.position().x(), mouseEvent.position().y());
-    m_mapView->identifyLayersAsync(mouseEvent.position(), tolerance, returnPopups).then(this, [this](const QList<IdentifyLayerResult*>& results)
+    auto f = m_mapView->identifyLayersAsync(mouseEvent.position(), tolerance, returnPopups).then(this, [this](const QList<IdentifyLayerResult*>& results)
```
bri12415, line 193: In some places we store `f` and pass it to `addTask`, and in others we just pass the future in place with the result of the function call (like on line 448). I'm wondering if readability would improve or suffer if we just do this one way instead of both, what do you think?

SampleViewer/TaskCanceler.h
```diff
@@ -0,0 +1,74 @@
...
+    for (auto& watcher : m_watchers)
+    {
+      if (watcher->isRunning())
+      {
+        watcher->cancel();
+      }
+    }
+  }
+
+  std::unordered_set<std::shared_ptr<QFutureWatcherBase>> m_watchers;
```
bri12415, line 69: Are we able to use unique_ptrs here instead of shared? I might be wrong in this, but I don't see the elements of `m_watchers` being needed anywhere else?

bri12415:
> This is overkill to deploy to every sample, but it does provide a

I think your write-up may have been cut off

JamesMBallard:
Whoops, I finished off my write up.

CppSamples/UtilityNetwork/TraceUtilityNetwork/TraceUtilityNetwork.cpp
JamesMBallard: My intent was to try to have this be transparent and not bloat or complicate the sample workflow(s). We can do it either way, so I will standardize it for consistency.

SampleViewer/TaskCanceler.h
JamesMBallard: It captures the shared_ptr in the `&QFutureWatcher<T>::finished` lambda so it can remove it without needing to do a secondary lookup.

I could change to std::unique_ptr, but it might need lots of tweaks. What do you think?

bri12415: Oh sorry I missed that part, thanks for clarifying. I think this is fine how it is then 👍

CppSamples/UtilityNetwork/PerformValveIsolationTrace/PerformValveIsolationTrace.cpp
```diff
@@ -252,10 +256,13 @@ void PerformValveIsolationTrace::performTrace()
...
-    {
-      onTraceCompleted_();
-    });
+
+    m_taskCanceler->addTask(
+      m_utilityNetwork->traceAsync(traceParameters).then(this, [this](QList<UtilityTraceResult*>)
+      {
+        onTraceCompleted_();
+      })
+    );
```
bri12415, line 265:
```suggestion
    m_taskCanceler->addTask(m_utilityNetwork->traceAsync(traceParameters).then(this, [this](QList<UtilityTraceResult*>)
    {
      onTraceCompleted_();
    }));
```

Nit pick, feel free to ignore if you want. This just seems to be how the other ones are formatted.

bri12415 (approved):

SampleViewer/TaskCanceler.h
```diff
@@ -0,0 +1,74 @@
...
+    watcher->setFuture(future);
+
+    std::lock_guard<std::mutex> lock(m_mutex);
+    m_watchers.insert(watcher);
+  }
+
+private:
+  void cancelAllTasks_()
+  {
+    std::lock_guard<std::mutex> lock(m_mutex);
```
jared-2016, line 59: This might be a silly question but, is it possible to not obtain ownership in this case causing this to not cancel all tasks upon destruction?

JamesMBallard: No. Any thread holding the lock is only momentarily modifying the `std::unordered_set`. So even if it's locked, it would be temporarily. Presumably new tasks could only be added on the main thread, and destruction would happen on the main thread as well so it shouldn't be possible (_in theory_) for a task to be added at the same time we're locking here and canceling all tasks.

Even if that happened, it wouldn't deadlock.

jared-2016: > Presumably new tasks could only be added on the main thread, and destruction would happen on the main thread as well so it shouldn't be possible (in theory) for a task to be added at the same time we're locking here and canceling all tasks.

this is what came to mind and prompted my question. Thanks for the clarification.

PR #1754, opened 2024-10-17
Tan11389/scene grid app update

tanneryould:
# Description

Updates and renames the "Display Grid" sample to "Show Grid" and adds 3D functionality.

## Type of change

**Platforms tested on**:

- [x] Windows
- [x] macOS


CppSamples/DisplayInformation/ShowGrid/ShowGrid.cpp
```diff
@@ -0,0 +1,248 @@
...
+  m_sceneView->setVisible(false);
+
+  emit viewChanged();
+}
+
+
+void ShowGrid::setViewType(const QString& viewType)
+{
+  m_currentViewType = viewType;
+  if (viewType == "MapView")
```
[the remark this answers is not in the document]
ldanzinger: Similar feedback for the other string comparison checks throughout (grid type, etc). I can see the fact that these are displayed in the UI adds complications, but it's worth thinking about

CppSamples/DisplayInformation/ShowGrid/README.metadata.json
```diff
@@ -25,13 +25,12 @@
...
         "MapView",
         "SimpleLineSymbol",
+        "SceneView",
         "TextSymbol",
         "USNGGrid",
         "UTMGrid"
     ],
-    "redirect_from": [
-        "/qt/latest/cpp/sample-code/sample-qt-displaygrid.htm"
-    ],
```
ldanzinger, line 34: ❓ do we want to remove this? presumably not. people or websites might have bookmarks or links to the old sample, and they should get redirected to the new one

CppSamples/DisplayInformation/ShowGrid/ShowGrid.qml
```diff
@@ -19,24 +24,116 @@ import QtQuick.Controls
...
 import Esri.Samples
 
-DisplayGridSample {
-    id: displayGrid
-    clip: true
-    width: 800
-    height: 600
-
-    property real centerWindowY: (displayGrid.height / 2) - (styleWindow.height / 2)
+Item {
```
ldanzinger, line 27: ❓ Typically, the *Sample is the root type. Is the Item being root necessary?

CppSamples/DisplayInformation/ShowGrid/ShowGrid.qml
```diff
@@ -19,24 +24,116 @@ import QtQuick.Controls
...
+                mapQuickView.forceActiveFocus();
+            else
+                sceneQuickView.forceActiveFocus();
+        }
+    }
+
+    Row {
+        id: viewButtonsRow
+        anchors
+        {
```
ldanzinger, line 64: 🙃 
```suggestion
        anchors {
```

CppSamples/DisplayInformation/ShowGrid/ShowGrid.qml
```diff
@@ -19,24 +24,116 @@ import QtQuick.Controls
...
+            top: parent.top
+            horizontalCenter: parent.horizontalCenter
+            topMargin: 5
+        }
+
+        spacing: 2
+
+        Rectangle {
+            id: mapViewButton
+            property bool selected: gridSample.currentViewType === "MapView"
```
ldanzinger, line 74: ☁️ all of these strings floating around is what makes me think an enum would be better

CppSamples/DisplayInformation/ShowGrid/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+  fmt.setVersion(3, 2);
+  QSurfaceFormat::setDefaultFormat(fmt);
+#endif
+
+  QGuiApplication app(argc, argv);
+  app.setApplicationName(QString("ShowGrid - C++"));
+
+  // Access to Esri location services requires an API key. This can be copied below or used as a command line argument.
+  const QString apiKey = QString("");
+  setAPIKey(app, apiKey);
```
ldanzinger, line 45: ❌ this API Key terminology all looks old. The samples and templates should've been updated to use the "access token" terminology

ldanzinger (changes requested):
Thanks @tanneryould. A few required changes and a few things to think about to improve it. A git mv would've been good to reduce churn, retain history, etc

jared-2016:
similar to what Lucas mentioned, I would request you possibly go back and `git mv`. Looks like everything is new rather than just an update. I'll do a review anyhow but just wanted to make the request.

CppSamples/DisplayInformation/ShowGrid/README.metadata.json
```diff
@@ -1,6 +1,6 @@
 {
-    "category": "Display information",
-    "description": "Display coordinate system grids including Latitude/Longitude, MGRS, UTM and USNG on a map view. Also, toggle label visibility and change the color of grid lines and grid labels.",
+    "category": "Show grid",
```
jared-2016, line 2: Is there a new category for `Show grid` ?

CppSamples/DisplayInformation/ShowGrid/ShowGrid.h
```diff
@@ -0,0 +1,105 @@
...
+  // Properties
+  QString m_currentViewType;
+  QString m_currentGridType;
+  QString m_currentLineColor;
+  QString m_currentLabelColor;
+  QString m_currentLabelPosition;
+  QString m_currentLabelFormat;
+
+  // Constants
+  QStringList m_viewTypes = {"MapView", "SceneView"};
```
jared-2016, line 97: can we make these all const?

CppSamples/DisplayInformation/ShowGrid/main.cpp
```diff
@@ -0,0 +1,96 @@
+// Copyright 2024 Esri.
```
jared-2016, line 1: Yeah this looks like it's using an outdated template. They no longer look like this since the access token update, https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/v.next/Templates/CppSampleTemplate/main.cpp.tmpl

CppSamples/DisplayInformation/ShowGrid/ShowGrid.cpp
```diff
@@ -0,0 +1,248 @@
...
+  }
+  else if (viewType == "SceneView")
+  {
+    m_sceneView->setViewpointAsync(m_mapView->currentViewpoint(ViewpointType::BoundingGeometry), 0);
+
+    m_mapView->setVisible(false);
+    m_sceneView->setVisible(true);
+
+    m_mapView->setGrid(nullptr);
+    m_sceneView->setGrid(m_grid);
```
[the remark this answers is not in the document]
tanneryould: Can you elaborate on what you're thinking here? If I were to apply the changes to a base GeoView, I'd still need to cast the base class to its inherited class before calling methods like "setVisible" and unless I changed a fair amount of the sample template structure, I'd still need methods like setMapView and setSceneView to receive the components created from QML.

CppSamples/DisplayInformation/ShowGrid/ShowGrid.cpp
```diff
@@ -0,0 +1,248 @@
...
+  m_grid->setVisible(visible);
+}
+
+void ShowGrid::setLabelsVisible(bool visible)
+{
+  m_grid->setLabelsVisible(visible);
+}
+
+void ShowGrid::setLineColor(const QString& lineColor)
+{
```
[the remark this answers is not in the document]
tanneryould: Not quite for all of them, because when we create a new grid, we apply all the current UI settings to it. So if the user has changed the grid to have blue lines and then we swap from a UTM grid to a USNG grid, we use the `setLineColor` method to overwrite the default grid properties and we want that to fire even if `m_lineColor == lineColor`, where m_lineColor is still reflecting what we have in the UI and lineColor is being supplied from the UI. I can check things like m_grid->isVisible() though because that's just a boolean, but things like colors and placement enums are just a bit arduous to check and will bloat the code. If I remember correctly, setting a property to itself is a no-op in RTC so an initial check wouldn't save much time either.

```cpp
  // Set properties from current UI values
  setGridVisible(m_gridVisible);
  setLabelsVisible(m_labelsVisible);
  setLineColor(m_currentLineColor);
  setLabelColor(m_currentLabelColor);
  setLabelPosition(m_currentLabelPosition);
  setLabelFormat(m_currentLabelFormat);
  ```

File: CppSamples/DisplayInformation/ShowGrid/README.metadata.json (hunk 7 above)
Comment (tanneryould) [reply CppSamples/DisplayInformation/ShowGrid/README.metadata.json:2]: Yeah, I'm not sure why the category got removed, sorry

File: CppSamples/DisplayInformation/ShowGrid/ShowGrid.qml (hunk 3 above)
Comment (tanneryould) [reply CppSamples/DisplayInformation/ShowGrid/ShowGrid.qml:27]: I'm not sure what you mean? Here's our most recent template: https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/v.next/Templates/CppSampleTemplate/sample.qml.tmpl#L21

**Comment by tanneryould** [comment]
A `git mv` would have been good, yeah -- back when I re did this sample in 200.5 however this was still using the old template style with the `void DisplayGrid::componentComplete()` way of doing things, so this really is a full delete of the old sample and addition of a new one with a different template but I can make some git changes.

**Comment by tanneryould** [comment]
@jared-2016 and @ldanzinger thank you a ton for your reviews so far. I believe I've been able to address most of your feedback but there were a few things I was a little unsure of. I've pushed up my changes and used "git mv" to track changes with the renamed files. If you want to see that specifically, you can view this specific commit: https://github.com/Esri/arcgis-maps-sdk-samples-qt/pull/1754/commits/08382f85af38483677c606da7ce2c3596e32e493

Thank you!

File: CppSamples/DisplayInformation/ShowGrid/ShowGrid.cpp [hunk 12]
```diff
@@ -0,0 +1,276 @@
...
+  else if (viewType == s_sceneView)
+  {
+    m_sceneView->setViewpointAsync(m_mapView->currentViewpoint(ViewpointType::BoundingGeometry), 0);
+
+    m_mapView->setVisible(false);
+    m_sceneView->setVisible(true);
+
+    m_mapView->setGrid(nullptr);
+    m_sceneView->setGrid(m_grid);
+  }
```
ldanzinger, line 145: I haven't built or tested this code, but I was thinking of something conceptually like this
```suggestion
  auto geoView = viewType == s_mapView ? m_mapView : m_sceneView;
  geoView->setViewpointAsync(geoView->currentViewpoint(ViewpointType::BoundingGeometry), 0);

  geoView->setVisible(true);
  geoView->setVisible(false);

  geoView->setGrid(nullptr);
  geoView->setGrid(m_grid);  
```

CppSamples/DisplayInformation/ShowGrid/README.metadata.json
ldanzinger: You're right, I confused myself with the component complete changes

CppSamples/DisplayInformation/ShowGrid/ShowGrid.cpp
tanneryould: GeoView doesn't have a `setVisible` method - that's only exposed to MapView and SceneView specifically. Otherwise I'd be able to do something like this:

```cpp
  auto geoView = viewType == s_mapView ? m_mapView : m_sceneView;
  auto oldGeoView = viewType != s_mapView ? m_mapView : m_sceneView;
  geoView->setViewpointAsync(oldGeoView->currentViewpoint(ViewpointType::BoundingGeometry), 0);
  geoView->setVisible(true);
  oldGeoView->setVisible(false);
  oldGeoView->setGrid(nullptr);
  geoView->setGrid(m_grid); 
  ```

ldanzinger: @tanneryould your call on which is better, just seemed like there was an opportunity to streamline

tanneryould: While I agree there might be a clever way to streamline it and that would be cool, I think that's a bit out of scope for this sample. I'd likely have to rewrite a bit of it and maybe somehow control each view's visibility on the qml side.

tanneryould: Solved, thank you!

CppSamples/DisplayInformation/ShowGrid/ShowGrid.qml
tanneryould: That's fair, but we only do two string comparisons on the front end and setting up enums for all (or some) of the properties might get a lot more cumbersome

ldanzinger (approved):
LGTM

ldanzinger:
@tanneryould looks like this change requires updates to the sample viewer imports and initialization

PR #1787, opened 2025-02-26
Remove extract_android_style entry from manifest file

bri12415:
# Description

In our provided `AndroidManifest.xml` file for the `SampleViewer`, we have this entry:
```
<meta-data android:name="android.app.extract_android_style" android:value="default"/>
```

But according to the [Qt documentation](https://doc.qt.io/qt-6/android-manifest-file-configuration.html#style-extraction), `default` is intended to be used with either quick controls 1 or widgets. We use [quickcontrols2](https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/v.next/SampleViewer/SampleViewer.pro#L22) in our sample viewer.

The [default value](https://doc.qt.io/qt-6/android-manifest-file-configuration.html#qt-specific-meta-data) for the `extract_android_style` is `minimal`, which corresponds to quick controls 2. So lets just remove the entry and use the default value.

## Type of change

- [x] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [x] Android
- [ ] Linux
- [ ] macOS
- [ ] iOS

### DIFF
--- a/SampleViewer/android/AndroidManifest.xml
+++ b/SampleViewer/android/AndroidManifest.xml
@@ -25,7 +25,6 @@
             </intent-filter>
             <meta-data android:name="android.app.lib_name" android:value="-- %%INSERT_APP_LIB_NAME%% --"/>
             <meta-data android:name="android.app.arguments" android:value="-- %%INSERT_APP_ARGUMENTS%% --"/>
-            <meta-data android:name="android.app.extract_android_style" android:value="default"/>
         </activity>
     </application>
 </manifest>


JamesMBallard (commented):
This is confusing. The "default" no longer works, and by removing the entry, we get the _new_ default, which is "minimal"? 😕 

I thought https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/v.next/SampleViewer/resources/qtquickcontrols2.conf file controlled the styling.

Also, do you know if this will change the app styling enough to invalidate the release team's image comparisons on Android?

bri12415:
> This is confusing. The "default" no longer works, and by removing the entry, we get the new default, which is "minimal"?

Yes, the wording is confusing. So the default value for `android.app.extract_android_style` is `minimal` if no value is provided. They also have a value for `android.app.extract_android_style` called `default`. According to their documentation, that `default` value is intended for use with Qt widgets and quickcontrols1. It says this for 6.8 LTS and 6.5 LTS.

> I thought https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/v.next/SampleViewer/resources/qtquickcontrols2.conf file controlled the styling.

I'm not sure what the interaction is between this conf file and that entry in the xml, but to be honest I wasn't aware of this conf file when I made this change. To me, it looks like the conf file is just for configuring quickcontrols2 while the entry in the manifest is for determining which styling system to use.

> Also, do you know if this will change the app styling enough to invalidate the release team's image comparisons on Android?

I don't know, but my guess would be yes. I can try and visually confirm that, but I'm guessing there would be differences that I wouldn't be able to pick up on with just my eyes.

PR #1802, opened 2025-04-14
Rename KML -> Kml

bri12415:
# Description

Rename `CreateKMLMultiTrack.qrc` -> `CreateKmlMultiTrack.qrc`. Case difference is causing qmlimportscanner errors because its referenced as its lowercase version in the .pro file. 
## Type of change

- [x] Bug fix
- [ ] Other enhancement


aji13187 (approved):
Thanks!

PR #1816, opened 2025-06-12
Run all continuations on the app's thread.

JamesMBallard:
This can cause problems if we don't when there's a need to set parents across thread on new objects.

This is primarily defensive changes. Not all of these are associated with any specific problems, but we want to encourage best practices and consistency, so it is best to run continuations on the app's thread and not directly on callbacks initiated from internal code on a QThread. In some situations setting parents can be problematic.

# Description

## Type of change

- [x] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS


jared-2016 (approved):
👍

PR #1851, opened 2025-09-09
Combine CRUD Samples in feature service

min14751:
# Description

## Type of change

- [ ] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [x] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [x] iOS

## Checklist

- [x] Self-review of changes
- [x] No unrelated changes have been made to any other code or project files
- [x] Code is commented with correct formatting (CTRL+i)
- [x] There is no leftover commented code
- [x] Screenshots are correct size and display in description tab (500px by 500px, platform agnostic)
- [x] If adding a new sample, it is added to the sample viewer


SampleViewer/mainSample.cpp
```diff
@@ -1,5 +1,18 @@
...
+// laws, treaties, and conventions.
+//
+// For additional information, contact:
+// Environmental Systems Research Institute, Inc.
+// Attn: Contracts and Legal Services Department
+// 380 New York Street
+// Redlands, California, 92373
+// USA
+//
+// email: contracts@esri.com
```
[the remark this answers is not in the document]
min14751: Should we change this in the mainSample.cpp file?

min14751:
This combined sample can be found under EditData->ManageFeaturesFeatureServiceSample

CppSamples/EditData/ManageFeaturesFeatureService/README.metadata.json
```diff
@@ -16,7 +16,7 @@
         "ServiceFeatureTable"
     ],
     "redirect_from": [
-        "/qt/latest/cpp/sample-code/sample-qt-addfeaturesfeatureservice.htm"
+        ""
```
ldanzinger: this should contain all the URLs from the old sample names. If users/docs link to the old URLs, we want them to redirect to the new one

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.qml
```diff
@@ -0,0 +1,218 @@
...
+        font.pixelSize: 12
+        text: qsTr("Update Attributes")
+    }
+
+    // Simple right panel - only takes space it needs
+    Rectangle {
+        id: rightPanel
+        anchors.top: parent.top
+        anchors.right: parent.right
+        anchors.margins: 10
```
jared-2016:
```suggestion
        anchors {
            top: parent.top
            right: parent.right
            margins: 10
        }
```

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.qml
```diff
@@ -0,0 +1,218 @@
...
+        color: palette.window
+        border.color: palette.windowText
+        border.width: 1
+        radius: 5
+
+        Column {
+            anchors.top: parent.top
+            anchors.left: parent.left
+            anchors.right: parent.right
+            anchors.margins: 10
```
jared-2016:
```suggestion
            anchors {
                top: parent.top
                left: parent.left
                right: parent.right
                margins: 10
            }
```

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.qml
```diff
@@ -0,0 +1,218 @@
...
+                width: parent.width
+                font.pixelSize: 11
+                wrapMode: Text.WordWrap
+                text: {
+                    switch(addFeaturesSample.operationMode) {
+                    case 0: return "Click anywhere on the map to add new features"
+                    case 1: return "Click on an existing feature to select and delete it"
+                    case 2: return "Click on an existing feature to select and update its attributes"
+                    case 3: return "Click to select a feature, then click again to move it to a new location"
+                    default: return ""
```
jared-2016:
```suggestion
                    switch(addFeaturesSample.operationMode) {
                    case 0:
                        return "Click anywhere on the map to add new features"
                    case 1:
                        return "Click on an existing feature to select and delete it"
                    case 2:
                        return "Click on an existing feature to select and update its attributes"
                    case 3:
                        return "Click to select a feature, then click again to move it to a new location"
                    default:
                        return ""
```

jared-2016: If it makes sense, maybe consider making `operationModes` an enum instead so the case statements are more readable.

Also does `qsTr` work for the strings?

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.h
```diff
@@ -14,59 +14,84 @@
...
-  explicit UpdateAttributesFeatureService(QQuickItem* parent = nullptr);
-  ~UpdateAttributesFeatureService() override;
+  enum OperationMode
+  {
+    AddFeatures = 0,
+    DeleteFeatures = 1,
+    UpdateAttributes = 2,
+    UpdateGeometry = 3
+  };
+  Q_ENUM(OperationMode)
```
jared-2016: Looks like you do have an enum here for OperationMode. Should use this instead and expose it as a QPROPERTY and register the enum with QML. We have a handful of examples throughout the samples and toolkit for reference, see below. Let me know if you have any questions about this.
https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/5d4e3a4a292746b0d950d593ac4751f38f5a87af/uitools/common/src/PopupElementViewItem.h#L40
https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/main/uitools/common/src/QmlEnums.h

You should also make this an enum class instead as it has stronger type safety, enforces scoped enumerators which avoids naming conflicts and improves code clarity, and is generally considered best practice for modern C++.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,541 @@
...
+      if (m_serviceGeodatabase->table(0))
+      {
+        ServiceFeatureTable* featureTable = m_serviceGeodatabase->table(0);
+        m_featureLayer = new FeatureLayer(featureTable, this);
+        m_map->operationalLayers()->append(m_featureLayer);
+
+        connectSignals();
+      }
+    }
+  });
```
jared-2016, line 113: nitpick: could make this a guard statement to reduce indentation. Up to you.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,541 @@
...
+  {
+    if (m_operationMode != AddFeatures)
+    {
+      m_featureLayer->clearSelection();
+      emit hideWindow();
+      if (m_operationMode == UpdateGeometry)
+      {
+        m_featureSelected = false;
+      }
+    }
```
jared-2016: nitpick: same comment as loadStatusChanged above, could check the inverse and exit  since all other operationModes advance which would reduce indentation.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+void ManageFeaturesFeatureService::deleteSelectedFeature()
+{
+  m_serviceGeodatabase->table(0)
+    ->deleteFeatureAsync(m_selectedFeature)
+    .then(this, [this]()
+  {
+    // Handle the completion of applyEditsAsync from the ServiceGeodatabase
+    m_serviceGeodatabase->applyEditsAsync().then(this, [](const QList<FeatureTableEditResult*>& tableEditResults)
+    {
```
jared-2016, line 191: styling question, I wonder why `deleteFeatureAsync` is broken down to new lines where `applyEditsAsync` is not. I would stick with one or the other.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,541 @@
...
+  m_operationMode = mode;
+
+  // Clear any existing selection when switching modes
+  if (m_featureLayer)
+  {
+    m_featureLayer->clearSelection();
+    m_featureSelected = false;
+  }
+  emit hideWindow();
+
```
jared-2016:
```suggestion
```

jared-2016 (commented):
I am realizing after i've finished my review that this was mostly an effort of consolidation of existing samples into one offering. Some of my suggestions might fall outside of the scope of this PR but i'll leave that for you to decide. There are some improvements that can be made and since we are here, now seems like a good time to do them.

SampleViewer/mainSample.cpp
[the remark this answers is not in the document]
jared-2016: weird, i imagine this was the githooks. I would stick with what we had before, https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/372b42e61b0519321431fcdcaecf65e290b5179f/SampleViewer/mainSample.cpp#L1-L14

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,541 @@
...
+void ManageFeaturesFeatureService::onApplyEditsCompleted_(const QList<FeatureTableEditResult*>& tableEditResults)
+{
+  if (tableEditResults.isEmpty())
+  {
+    qDebug() << "No table edit results";
+    return;
+  }
+
+  // Get the first table result
+  FeatureTableEditResult* tableResult = tableEditResults.first();
```
bri12415: I like your FeatureTableResultLock helper, can we use that here?

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,541 @@
...
+    // Handle the completion of applyEditsAsync from the ServiceGeodatabase
+    m_serviceGeodatabase->applyEditsAsync().then(this, [](const QList<FeatureTableEditResult*>& tableEditResults)
+    {
+      if (tableEditResults.isEmpty())
+      {
+        return;
+      }
+
+      // Get the first table result
+      FeatureTableEditResult* tableResult = tableEditResults.first();
```
bri12415: Can we use your FeatureTableResultLock helper here as well?

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,541 @@
...
+    m_serviceGeodatabase->table(0)
+      ->queryFeaturesAsync(queryParams)
+      .then(this, [this](FeatureQueryResult* featureQueryResult)
+    {
+      if (featureQueryResult && featureQueryResult->iterator().hasNext())
+      {
+        // first delete if not nullptr
+        if (m_selectedFeature)
+        {
+          delete m_selectedFeature;
```
bri12415: I'm not a fan of the raw `delete`, but I think that might be the pattern throughout our samples.

For member variables that aren't managed through Qt's parent system (like here), I personally would prefer a `std::unique_ptr`. What do you think?

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,541 @@
...
+    }
+  }
+
+  // Delete table edit results
+  qDeleteAll(tableEditResults);
+}
+
+void ManageFeaturesFeatureService::handleAddFeatureClick_(QMouseEvent& mouseEvent)
+{
+  // ADD FEATURES MODE
```
bri12415: This is a nitpick, but I think these all caps comments on these next few handler functions may be overkill given they essentially repeat the function name (other than "click" vs "mode"). If you disagree I am fine with them staying.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.qml
```diff
@@ -0,0 +1,218 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// [Legal]
+
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+import Esri.Samples
+import Esri.ArcGISRuntime.Toolkit
```
bri12415, line 21: I think we usually provide some default toolkit importing code in the `.pro` file when using the toolkit in our samples.

[ChangeBasemap](https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/ede6beee7909f07271736385006b7cdde3cfd150/CppSamples/Maps/ChangeBasemap/ChangeBasemap.pro#L33-L40)

Can you add that to this sample's `.pro` file?

CppSamples/EditData/ManageFeaturesFeatureService/main.cpp
```diff
@@ -62,7 +62,7 @@ int main(int argc, char *argv[])
   }
 
   // Initialize the sample
-  AddFeaturesFeatureService::init();
+  ManageFeaturesFeatureService::init();
 
   // Initialize application view
   QQmlApplicationEngine engine;
```
bri12415, line 71: I think we'll need a couple more things in this main file to account for our `toolkit` usage.

[ChangeBasemap](https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/ede6beee7909f07271736385006b7cdde3cfd150/CppSamples/Maps/ChangeBasemap/main.cpp#L76-L77)

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
min14751: not entirely sure on `std::unique_ptr` since `ArcGISFeature` is a `Q_Object`. We can avoid the raw delete by using a `QPointer` and since we are doing `m_selectedFeature->setParent(this);` deleting and making the pointer `null` should be taken care of when the parent is cleared

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.qml
min14751: Updated. I think some of the older samples are missing the Toolkit boilerplate.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
min14751: Not sure why but our `command + i` formatting forces it to be like this.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
bri12415: > We can avoid the raw delete by using a QPointer

I don't believe `QPointer` alone handles lifetime management of the object it holds, I believe it's just a `weak_ptr` designed for QObjects. But like you said, setting the parent will handle the lifetime. I didn't see that happening on the line below, my mistake. 

I do still think it's confusing to both set the parent and manually delete the object, that seems like two lifetime management systems being mingled. But that's just my opinion and can be wrong. If you want to leave it this way that's fine.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,530 @@
...
+      Feature* feature = static_cast<Feature*>(geoElement);
+
+      // Select the feature
+      QueryParameters query;
+      query.setObjectIds(QList<qint64>{feature->attributes()->attributeValue(QStringLiteral("objectid")).toLongLong()});
+      (void)m_featureLayer->selectFeaturesAsync(query, SelectionMode::New);
+
+      // Set selected feature member and switch to selected mode
+      m_selectedFeature = static_cast<ArcGISFeature*>(feature);
+      m_selectedFeature->setParent(this);
```
min14751: `m_selectedFeature` is now a QPointer so we only set its parent and dont call delete.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
JamesMBallard: >I do still think it's confusing to both set the parent and manually delete the object, that seems like two lifetime management systems being mingled

I feel strongly about this. We should not mix the two systems. Please pick either QObject parent ownership, or some alternative smart pointer ownership, but not both.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.qml
```diff
@@ -0,0 +1,228 @@
...
+            }
+
+            Label {
+                id: instructionText
+                width: parent.width
+                font.pixelSize: 11
+                wrapMode: Text.WordWrap
+                text: {
+                    switch(addFeaturesSample.operationMode) {
+                    case ManageFeaturesFeatureService.OperationMode.AddFeatures:
```
jared-2016, line 118: much nicer IMO.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.qml
```diff
@@ -0,0 +1,228 @@
...
+            calloutData: view.calloutData
+            leaderPosition: Callout.LeaderPosition.Automatic
+            // Override the background to keep it consistent with button images
+            background: Rectangle {
+                color: "#ffffff"
+                border.color: "#000000"
+            }
+            
+            // Delete mode - show trash button, Update mode - show accessory button
+            accessoryButtonType: addFeaturesSample.operationMode === 1 ? "Custom" : (addFeaturesSample.operationMode === 2 ? "Info" : "None")
```
jared-2016:
```suggestion
            accessoryButtonType: addFeaturesSample.operationMode === ManageFeaturesFeatureService.OperationMode.AddFeatures ? "Custom" : (addFeaturesSample.operationMode === ManageFeaturesFeatureService.OperationMode.DeleteFeatures ? "Info" : "None")
```

There are still a few more places we can replace `int` with the enum you used below. I won't comment on all of them but anywhere there is `addFeaturesSample.operationMode` we should use the enum instead of the int. I put in the suggested change but you might want to break this ternary onto multiple lines so it doesn't run on so long.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
jared-2016: 🤷

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.qml
min14751: Yes, i agree.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.h
```diff
@@ -14,59 +14,84 @@
...
 private:
-  void onIdentifyLayerCompleted_(Esri::ArcGISRuntime::IdentifyLayerResult* identifyResult);
-  void onQueryFeaturesCompleted_(Esri::ArcGISRuntime::FeatureQueryResult* featureQueryResult);
-  void onApplyEditsCompleted_(const QList<Esri::ArcGISRuntime::FeatureEditResult*>& featureEditResults);
   void connectSignals();
-  QString featureType() const;
+  void onIdentifyLayerCompleted_(Esri::ArcGISRuntime::IdentifyLayerResult* identifyResult);
+  void onApplyEditsCompleted_(const QList<Esri::ArcGISRuntime::FeatureTableEditResult*>& tableEditResults);
+
+  void handleAddFeatureClick_(QMouseEvent& mouseEvent);
```
jared-2016: can these be const?
```suggestion
  void handleAddFeatureClick_(const QMouseEvent& mouseEvent);
```

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.h
```diff
@@ -14,59 +14,84 @@
...
-  explicit UpdateAttributesFeatureService(QQuickItem* parent = nullptr);
-  ~UpdateAttributesFeatureService() override;
+  explicit ManageFeaturesFeatureService(QQuickItem* parent = nullptr);
+  ~ManageFeaturesFeatureService() override;
 
   void componentComplete() override;
   static void init();
+
+  Q_INVOKABLE void deleteSelectedFeature();
   Q_INVOKABLE void updateSelectedFeature(QString fieldVal);
```
jared-2016: same const question for this as well

jared-2016 (commented):
Looking good. A few more suggested changes.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.h
min14751: Cannot be const beacuse we are updating `m_serviceGeodatabase`

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,530 @@
...
+
+    if (!identifyResult->geoElements().isEmpty())
+    {
+      GeoElement* geoElement = identifyResult->geoElements().at(0);
+      Feature* feature = static_cast<Feature*>(geoElement);
+
+      // Select the feature
+      QueryParameters query;
+      query.setObjectIds(QList<qint64>{feature->attributes()->attributeValue(QStringLiteral("objectid")).toLongLong()});
+      (void)m_featureLayer->selectFeaturesAsync(query, SelectionMode::New);
```
bri12415: I think as a rule we avoid c-syle casts. Why do we need this cast here?

min14751: Got it. I used the cast there because i was getting a warning  `Ignoring return value of function declared with 'nodiscard' attribute .`

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.h
jared-2016: apologies for the confusion const ref the parameter
```suggestion
  Q_INVOKABLE void updateSelectedFeature(const QString& fieldVal);
```

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,530 @@
...
+
+    if (!identifyResult->geoElements().isEmpty())
+    {
+      GeoElement* geoElement = identifyResult->geoElements().at(0);
+      Feature* feature = static_cast<Feature*>(geoElement);
+
+      // Select the feature
+      QueryParameters query;
+      query.setObjectIds(QList<qint64>{feature->attributes()->attributeValue(QStringLiteral("objectid")).toLongLong()});
+      m_featureLayer->selectFeaturesAsync(query, SelectionMode::New);
```
jared-2016: I think the reason why you are getting the `nodiscard` warning is because this will return a result `FeatureQueryResult*` which if it's lifetime is not handled will live as long as it's calling parent, `m_featureLayer` in this case. So even though you don't want to do anything with the result, you should at least make sure we don't create a bunch of these that live on until the application dies, I think we call this a space leak.

For an example of how we usually handle getting a result that we don't want to do anything with in this case, https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/ede6beee7909f07271736385006b7cdde3cfd150/CppSamples/UtilityNetwork/ValidateUtilityNetworkTopology/ValidateUtilityNetworkTopology.cpp#L539-L542

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,534 @@
...
+  connect(m_serviceGeodatabase, &ServiceGeodatabase::loadStatusChanged, this, [this](LoadStatus loadStatus)
+  {
+    if (loadStatus != LoadStatus::Loaded)
+    {
+      return;
+    }
+    // Get the first table(id 0) from the ServiceGeodatabase
+    if (m_serviceGeodatabase->table(0))
+    {
+      ServiceFeatureTable* featureTable = m_serviceGeodatabase->table(0);
```
JamesMBallard:
```suggestion
    if (ServiceFeatureTable* table = m_serviceGeodatabase->table(0); table)
    {
      ServiceFeatureTable* featureTable = table;
```

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    ~FeatureTableResultLock()
+    {
+      qDeleteAll(results);
+    }
+
+    const QList<FeatureTableEditResult*>& results;
+  };
+} // namespace
```
JamesMBallard: This is optional, but I stumbled upon QScopeGuard recently. It can do all of this for you.

https://doc.qt.io/qt-6/qscopeguard.html

Usage would be:
```c++
const QScopeGuard deleteAllOnExit{[&tableEditResults]()
{
  qDeleteAll(tableEditResults);
});
```

Warning - code above is untested.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,534 @@
...
+  });
+}
+
+void ManageFeaturesFeatureService::updateSelectedFeature(const QString& fieldVal)
+{
+  // If the last connection is still hanging around we want to ensure it is disconnected.
+  disconnect(m_featureLoadStatusChangedConnection);
+
+  // Connect to load status changed signal, remember the connection so we can kill it once the slot has invoked.
+  m_featureLoadStatusChangedConnection = connect(m_selectedFeature, &ArcGISFeature::loadStatusChanged, this, [this, fieldVal](LoadStatus)
```
JamesMBallard: Optional since your logic works.

Another way to think about this is that you are disconnecting and reconnecting this repeatedly solely because the `fieldVal` capture changes. You could consider storing that to a class member and then the connection remains.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,534 @@
...
+  {
+    if (m_selectedFeature->loadStatus() == LoadStatus::Loaded)
+    {
```
JamesMBallard: Two recommendations here.

1. Early exit strategy. Do all your precondition checks and return if they're not met. That way you don't end up with this:

```C++
if (conditionA)
{
  if (conditionB)
  {
    if (conditionC)
    {
       // the important logic here we want to showcase
    }
  }
}
```
but rather:
```c++
if (!conditionA || !conditionB || !conditionC)
{
  return;
}

// the important logic here we want to showcase
```

It reduces code nesting.

2. Prefer `doneLoading` to `loadStatusChanged`.

If your logic is doing this `if (m_selectedFeature->loadStatus() == LoadStatus::Loaded)`, then you probably just need the `doneLoading` signal and you can check for any load error.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,534 @@
...
+          onApplyEditsCompleted_(tableEditResults);
+        });
+      });
+    }
+  });
+
+  // Check if feature is loaded
+  if (m_selectedFeature->loadStatus() == LoadStatus::Loaded)
+  {
+    // The connection is invoked so we now forget all about this connection after this point.
```
JamesMBallard: The connection management feels brittle. What are you trying to achieve?

If you want the connection to fire one time only, you can use [Qt::SingleShotConnection](https://doc.qt.io/qt-6/qt.html#ConnectionType-enum). Maybe there is more to it.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,534 @@
...
+}
+
+void ManageFeaturesFeatureService::handleIdentifyFeatureClick_(const QMouseEvent& mouseEvent)
+{
+  // Identify and select feature mode (for delete and update attributes)
+  // First clear the selection
+  m_featureLayer->clearSelection();
+
+  // Set the properties for qml
+  emit hideWindow();
```
JamesMBallard, line 408: This doesn't feel right. hideWindow is an action. Signals are usually state change events. If this was already like this, you can keep it this way. If this is new, I suggest changing it.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,534 @@
...
+        std::unique_ptr<FeatureQueryResult>{rawResult};
+      });
+
+      // Set selected feature member and switch to selected mode
+      m_selectedFeature = static_cast<ArcGISFeature*>(feature);
+      m_selectedFeature->setParent(this);
+      m_featureSelected = true;
+    }
+
+    delete identifyResult;
```
JamesMBallard: If the intention is to delete identifyResult at the end of the block, then just stick it into a unique_ptr. That way if `m_featureLayer->selectFeaturesAsync` fails, or anything else goes wrong, you are guaranteed to not leak.

identifyLayerAsync takes 6 args, and the sixth one (optional parent) is missing, so there is no confusion about ownership with this result. Nobody owns it, so much better to use unique_ptr.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.h
```diff
@@ -0,0 +1,97 @@
...
+  Q_INVOKABLE void updateSelectedFeature(const QString& fieldVal);
+
+  OperationMode operationMode() const;
+  void setOperationMode(OperationMode mode);
+  QString featureType() const;
+
+signals:
+  void featureSelected();
+  void featureTypeChanged();
+  void hideWindow();
```
JamesMBallard, line 72: Review this. It does not follow typical signal conventions.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.h
```diff
@@ -0,0 +1,97 @@
...
+  void operationModeChanged();
+
+private:
+  void connectSignals();
```
JamesMBallard: For consistency:

```suggestion
  void connectSignals_();
```

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.pro
```diff
@@ -1,5 +1,5 @@
 #-------------------------------------------------
-# Copyright 2015 Esri.
+# Copyright 2022 Esri.
```
JamesMBallard: May as well go with 2025.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
min14751: Implemented this. Much cleaner.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,507 @@
...
+      {
+        return;
+      }
+
+      // Get the first table result
+      FeatureTableEditResult* const tableResult = tableEditResults.first();
+      if (!tableResult->editResults().isEmpty())
+      {
+        // Get the first feature edit result from the table
+        FeatureEditResult* const featureEditResult = tableResult->editResults().first();
```
jared-2016:
```suggestion
        const FeatureEditResult* featureEditResult = tableResult->editResults().first();
```

jared-2016: typically if a variable is const we prefer it infront of the type.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.h
JamesMBallard: I see now this was already like this. You didn't add this signal. It can stay as it is.

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
min14751: Used Qt::SingleShotConnection

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
min14751: Existing code

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.h
min14751: existing code

CppSamples/EditData/ManageFeaturesFeatureService/ManageFeaturesFeatureService.cpp
```diff
@@ -0,0 +1,507 @@
...
+    {
+      return;
+    }
+
+    if (identifyResult->geoElements().isEmpty())
+    {
+      return;
+    }
+
+    Feature* const feature = static_cast<Feature*>(identifyResult->geoElements().at(0));
```
jared-2016:
```suggestion
    const Feature* feature = static_cast<Feature*>(identifyResult->geoElements().at(0));
```

jared-2016 (approved):
Sorry I didn't catch this one earlier but you've addressed all my feedback so marking approved even with the recommended change.

JamesMBallard (approved):

PR #1859, opened 2025-09-18
James/indoors sample updates

JamesMBallard:
# Description

Update the Indoors sample to use the correct IndoorsLocationDataSource constructor. This required some coordination because the sample already has structured logic to request permissions in sequence, and then to create the IndoorsLocationDataSource. Well now we must get the indoorsPositioningDefinition from the map, which means the map must be loaded. Now there are a few things to coordinate.

Conceptually, the sample must have all of bluetooth permissions, location permissions, and the map fully loaded before it can create the IndoorsLocationDataSource.

Important note - I cannot get the sample to actually start the data source on iOS. I compared against v.next and the same thing happens, so it doesn't appear to be any new bug I've introduced. I also added some logging statements and it does reach the new logic to create and load the data source.

## Type of change

- [ ] Bug fix
- [X] Sample viewer enhancement
- [ ] Other enhancement

**Platforms tested on**:

- [X] iOS


aji13187 (approved):
LGTM

CppSamples/Maps/ShowDeviceLocationUsingIndoorPositioning/ShowDeviceLocationUsingIndoorPositioning.cpp
```diff
@@ -116,46 +120,69 @@ void ShowDeviceLocationUsingIndoorPositioning::requestLocationPermissionThenSetu
...
+    trySetupIndoorsLocationDataSource();
   });
 }
 
-void ShowDeviceLocationUsingIndoorPositioning::checkPermissions()
+void ShowDeviceLocationUsingIndoorPositioning::validatePermissions()
 {
-  if (qApp->checkPermission(QBluetoothPermission{}) == Qt::PermissionStatus::Denied)
+  bool bluetoothPermissionGranted = false;
+  if (auto status = qApp->checkPermission(QBluetoothPermission{}); status == Qt::PermissionStatus::Denied)
```
min14751: is `auto` okay to use here?

CppSamples/Maps/ShowDeviceLocationUsingIndoorPositioning/ShowDeviceLocationUsingIndoorPositioning.cpp
```diff
@@ -116,46 +120,69 @@ void ShowDeviceLocationUsingIndoorPositioning::requestLocationPermissionThenSetu
...
+  {
+    bluetoothPermissionGranted = true;
+  }
 
+  bool locationPermissionGranted = false;
   QLocationPermission locationPermission{};
   locationPermission.setAccuracy(QLocationPermission::Accuracy::Precise);
   locationPermission.setAvailability(QLocationPermission::Availability::WhenInUse);
-  if (qApp->checkPermission(locationPermission) == Qt::PermissionStatus::Denied)
+  if (auto status = qApp->checkPermission(locationPermission); status == Qt::PermissionStatus::Denied)
```
min14751: same `auto` usage

JamesMBallard:
@min14751 I updated the auto usages. Good catch. Ready for review again.

min14751 (approved):
LGTM

PR #1866, opened 2025-09-30
Add overall percentage to ZipHelper

dtedder:
# Description

Adding properties and signals to allow tracking of the overall extraction progress for the archive.

## Type of change

- [ ] Bug fix
- [x] Other enhancement

**Platforms tested on**:

- [x] Windows
- [x] Android
- [ ] Linux
- [x] macOS
- [x] iOS

## Checklist

- [ ] Self-review of changes

### DIFF
--- a/SampleViewer/ZipHelper.cpp
+++ b/SampleViewer/ZipHelper.cpp
@@ -22,6 +22,7 @@
 
 // enable logging
 //#define ZIP_LOGGING_ENABLED
+static constexpr bool TOTAL_PROGRESS_ENABLED = true;
 
 #ifdef Q_OS_WIN
 #include <sys/utime.h>
@@ -156,6 +157,16 @@ bool ZipHelper::extractCurrentFile(const QString& outputFilePath)
 
             emit extractProgress(fileName, outputFilePath, percent);
 
+            if constexpr (TOTAL_PROGRESS_ENABLED)
+            {
+                // check for a new percentage but only by whole numbers
+                auto oldPercentTotal = static_cast<int>(percentTotal());
+                m_bytesExtracted += result;
+                auto newPercentTotal = static_cast<int>(percentTotal());
+                if (newPercentTotal != oldPercentTotal)
+                    emit extractProgressTotal(newPercentTotal);
+            }
+
             outputFile.write(buffer.data(), result);
             readBytes += result;
         }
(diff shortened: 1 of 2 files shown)


SampleViewer/ZipHelper.cpp
```diff
@@ -22,6 +22,7 @@
 
 // enable logging
 //#define ZIP_LOGGING_ENABLED
+static constexpr bool TOTAL_PROGRESS_ENABLED = false;
```
JamesMBallard: Is this primarily off because you don't want to cause any side effects or extensive testing?

Unless you have any concerns that is undesirable, I would keep it on. I think it's worse to have a new signal  `extractProgressTotal` that doesn't work and someone will have to dig into the code to figure out why.

dtedder: Mainly just because we don't have anything using it yet. Probably not a huge concern. But there is a cost associated with enabling this. We have to make an additional pass over all the files in the archive in order to compute the total size up front and the percentage itself is recalculated quite frequently.

I was also trying to avoid changing any signatures of existing methods on the class. But if that's acceptable, I can propose an alternative approach. ZipHelper is really only used in the SampleManager itself. So updating existing uses will be pretty manageable.

SampleViewer/ZipHelper.cpp
```diff
@@ -182,6 +192,19 @@ bool ZipHelper::extractCurrentFile(const QString& outputFilePath)
 bool ZipHelper::extractAll(QDir& outputDir)
 {
     bool haveFile = gotoFirstFile();
+
+    if (m_trackTotalProgress)
```
bri12415, line 194: Do you think we can combine this into the while loop below on line 208? I see we're iterating over `while (haveFile)`  here and below, might be simpler. Up to you, I don't feel strongly on it.

JamesMBallard (approved):

bri12415 (approved):

SampleViewer/ZipHelper.cpp
dtedder: The reason for the additional loop here is to quickly go through the entire archive and accumulate the total uncompressed size so that an _overall_ percentage extracted can be computed each time bytes are read into the buffer. I don't think there's a way to avoid this without adding a bit more complexity to ZipHelper's methods or making new ones on the zlib directly.

I pushed up another commit though to clean up the code a bit where these iterations are performed. A for-loop is a bit more concise and actually fits a bit more nicely I would say. We have explicit start/condition/increment statements already present in the class that can be used.

PR #1879, opened 2025-10-21
Calcite changes - Display information category

min14751:
# Description
Style changes for samples in the display information category. The changes replace hardcoded colors with colors from the `palette`. Changes have also been made to use qsTr() where applicable.

## Type of change

- [ ] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

## Checklist

- [x] Self-review of changes

### DIFF
--- a/CppSamples/DisplayInformation/BuildLegend/BuildLegend.qml
+++ b/CppSamples/DisplayInformation/BuildLegend/BuildLegend.qml
@@ -137,7 +137,7 @@ BuildLegendSample {
                         Label {
                             anchors.verticalCenter: parent.verticalCenter
                             width: 125
-                            text: qsTr(name)
+                            text: name
                             wrapMode: Text.WordWrap
                             font.pixelSize: 12
                         }
@@ -155,7 +155,7 @@ BuildLegendSample {
                         color: palette.highlight
 
                         Label {
-                            text: qsTr(section)
+                            text: section
                             font.bold: true
                             font.pixelSize: 13
                         }
--- a/CppSamples/DisplayInformation/BuildLegend/BuildLegend.qml
+++ b/CppSamples/DisplayInformation/BuildLegend/BuildLegend.qml
@@ -45,7 +45,7 @@ BuildLegendSample {
         property bool expanded: true
         height: 200
         width: 175
-        color: "lightgrey"
+        color: palette.base
         opacity: 0.95
         radius: 10
         clip: true
@@ -80,7 +80,7 @@ BuildLegendSample {
             Row {
                 spacing: 55
 
-                Text {
+                Label {
                     text: qsTr("Legend")
                     font {
                         pixelSize: 18
@@ -134,10 +134,10 @@ BuildLegendSample {
                             height: width
                             source: symbolUrl
                         }
-                        Text {
+                        Label {
                             anchors.verticalCenter: parent.verticalCenter
                             width: 125
-                            text: name
+                            text: qsTr(name)
                             wrapMode: Text.WordWrap
                             font.pixelSize: 12
                         }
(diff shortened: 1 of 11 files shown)


CppSamples/DisplayInformation/BuildLegend/BuildLegend.qml
```diff
@@ -134,10 +134,10 @@ BuildLegendSample {
                             height: width
                             source: symbolUrl
                         }
-                        Text {
+                        Label {
                             anchors.verticalCenter: parent.verticalCenter
                             width: 125
-                            text: name
+                            text: qsTr(name)
```
ldanzinger: qsTr should only be needed when wrapping a string literal (e.g. qsTr("name")). The purpose of it is for string translations, so in this specific case, it wouldn't have any use. I suggest removing it

CppSamples/DisplayInformation/ControlAnnotationSublayerVisibility/ControlAnnotationSublayerVisibility.qml
```diff
@@ -38,23 +38,24 @@ Item {
                 margins: 2
             }
             width: childrenRect.width
-            height: childrenRect.height
-            color: "white"
-            opacity: .75
+            height: childrenRect.height + 16
+            color: palette.base
```
ldanzinger, line 42: I notice this is all still inside of a Rectangle. Should it be a Pane? Is there a rule of thumb that we should be following now, for example, always use Pane instead of Rectangle (and same for Text and Label)? Or is it more nuanced than that?

CppSamples/DisplayInformation/BuildLegend/BuildLegend.qml
ldanzinger: this comment applies to several spots in this PR

min14751: got it

CppSamples/DisplayInformation/ControlAnnotationSublayerVisibility/ControlAnnotationSublayerVisibility.qml
min14751: Since we now use the `palette` we could do either or . Our current implementation mostly uses `Rectangle` and switching to `Pane`  would require more changes to get the alignment right. 
In terms of behavior `Pane` needs to be the parent as ive found it blocking clicks. We also use `radius` which is not available for a pane unless we override the rectangle it contains.

ldanzinger: Ok. If it looks good with both, I'm ok with using both, but we should consider if one is better than the other and potentially standardize. Pane blocking clicks sounds ideal - we often have verifications bounce back because you can still pan the map behind a rectangle, for example. And I'm sure the radius isn't really a requirement, whoever wrote it just thought it looked good. But if we are using calcite styling and can depend on types that we know look good, that's a more consistent experience and rule of thumb to follow. If you think this has merit, we can log a follow up issue to go back through and make more of a global change, after the calcite integration is complete

ldanzinger (approved):

CppSamples/DisplayInformation/ControlAnnotationSublayerVisibility/ControlAnnotationSublayerVisibility.qml
min14751: Yes we could revisit the samples after the calcite integration is complete to switch to pane where required. Definitely looks like a more appropriate and cleaner component to be using.

jared-2016 (approved):

PR #1882, opened 2025-10-23
Calcite changes - Routing category

min14751:
# Description

Styling changes for samples in the Routing category. Removed hardcoded colors to support both light and dark theme. Button changes and qsTr usage are part of this pr.

## Type of change

- [ ] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

## Checklist

- [ ] Self-review of changes


CppSamples/Routing/DisplayRouteLayer/DisplayRouteLayer.qml
```diff
@@ -51,7 +51,7 @@ Item {
             focus: true
             contentItem: Controls.ScrollView {
                 contentWidth: parent.width - 30
-                Text {
+                Controls.Label {
```
ldanzinger: weird, I wonder why we scope the import like we do... That seems unnecessary. I would try and remove the `as Controls` in the QtQuick.Controls import and see if we can remove that complexity. Seems unnecessary from what I can tell

ldanzinger: Oh I see, it's probably because we have a QtQuick.Controls type that is Popup, but our SDK has a type called Popup. Either way, there shouldn't be any conflicts, we don't register Popup as a QML type, so I don't think that will cause any problems

CppSamples/Routing/FindServiceAreasForMultipleFacilities/FindServiceAreasForMultipleFacilities.qml
```diff
@@ -57,7 +58,7 @@ Item {
 
     BusyIndicator {
         anchors.centerIn: parent
-        visible: true
+        visible: model.taskRunning
```
ldanzinger, line 61: assuming you tested this? seems like opportunity for a subtle bug to arise otherwise

ldanzinger (approved):
LGTM, some minor cleanup to consider, but you can address at your discretion

CppSamples/Routing/FindServiceAreasForMultipleFacilities/FindServiceAreasForMultipleFacilities.qml
min14751: With the calcite implementation of the `BusyIndicator` component we do not account for visibility so we have to set it ourselves to prevent it from being visible when its not running. 

The default Qt component does it with `opacity: control.running ? 1 : 0`. Both approaches have the same effect visually.

I could update the calcite implementation to follow this so we wouldn't have to set visibility.

CppSamples/Routing/DisplayRouteLayer/DisplayRouteLayer.qml
min14751: Removed scoped import and tested it. No problems as expected.

CppSamples/Routing/FindServiceAreasForMultipleFacilities/FindServiceAreasForMultipleFacilities.qml
ldanzinger: I would not change the toolkit right now, that could have further impacts we would need to test thoroughly. Better to keep this contained to the sample for now

jared-2016 (approved):

PR #1918, opened 2025-11-18
Sync map and scene sample fix

jared-2016:
# Description

Just checking the isNavigating property before trying to update the adjacent view to avoid the stuttering. Only one will be 'true` at a given time.

## Type of change

- [x] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

## Checklist

- [x] Self-review of changes

### DIFF
--- a/CppSamples/Scenes/SyncMapViewSceneView/SyncMapViewSceneView.cpp
+++ b/CppSamples/Scenes/SyncMapViewSceneView/SyncMapViewSceneView.cpp
@@ -81,6 +81,13 @@ void SyncMapViewSceneView::setSceneView(SceneQuickView* sceneView)
       m_mapView->setViewpointAsync(m_sceneView->currentViewpoint(ViewpointType::CenterAndScale), 0);
     }
   });
+  connect(m_sceneView, &SceneQuickView::mouseWheelChanged, this, [this]
+  {
+    if (m_mapView)
+    {
+      m_mapView->setViewpointAsync(m_sceneView->currentViewpoint(ViewpointType::CenterAndScale), 0);
+    }
+  });
 
   emit sceneViewChanged();
 }
@@ -110,6 +117,14 @@ void SyncMapViewSceneView::setMapView(MapQuickView* mapView)
     }
   });
   connect(m_mapView, &MapQuickView::mousePressed, this, [this]
+  {
+    if (m_sceneView)
+    {
+      m_sceneView->setViewpointAsync(m_mapView->currentViewpoint(ViewpointType::CenterAndScale), 0);
+    }
+  });
+
+  connect(m_mapView, &MapQuickView::mouseWheelChanged, this, [this]
   {
     if (m_sceneView)
     {
--- a/CppSamples/Scenes/SyncMapViewSceneView/SyncMapViewSceneView.cpp
+++ b/CppSamples/Scenes/SyncMapViewSceneView/SyncMapViewSceneView.cpp
@@ -74,6 +74,13 @@ void SyncMapViewSceneView::setSceneView(SceneQuickView* sceneView)
       m_mapView->setViewpointAsync(m_sceneView->currentViewpoint(ViewpointType::CenterAndScale), 0);
     }
   });
+  connect(m_sceneView, &SceneQuickView::mousePressed, this, [this]
+  {
+    if (m_mapView)
+    {
+      m_mapView->setViewpointAsync(m_sceneView->currentViewpoint(ViewpointType::CenterAndScale), 0);
+    }
+  });
 
   emit sceneViewChanged();
 }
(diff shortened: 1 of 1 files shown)


CppSamples/Scenes/SyncMapViewSceneView/SyncMapViewSceneView.cpp
```diff
@@ -67,11 +67,12 @@ void SyncMapViewSceneView::setSceneView(SceneQuickView* sceneView)
   m_sceneView = sceneView;
   m_sceneView->setArcGISScene(m_scene);
 
-  connect(m_sceneView, &SceneQuickView::viewpointChanged, this,
-          [this]
+  connect(m_sceneView, &SceneQuickView::viewpointChanged, this, [this]
   {
-    if (m_mapView && m_sceneView->isNavigating())
+    if (m_mapView && m_sceneView->isNavigating() && !m_mapView->isNavigating())
```
jared-2016, line 72: Logical operators are evaluated left to right so there is no risk of m_mapView being null and trying to access it's property at the end of the evaluation, https://en.cppreference.com/w/cpp/language/operator_precedence.html.

But looking at this, maybe it's worth just moving the last one into it's own if statement? I could go either way.

jared-2016: same obviously applies to MapQuickView::viewpointChanged below.

JamesMBallard: This is a strange check to begin with. It checks to be sure that `m_mapView` is not null, but assumes that `m_sceneView` cannot be null. I think you can leave those as they were.

Also, I don't know if `!m_mapView->isNavigating()` is bulletproof for this, but it might be. The moment both of them start navigating you can no longer tell who's driving the interactions and which one should just follow. Is this sufficient if you jump back and forth between the two and pan and zoom on each one? Ideally whichever one you interacted with last takes over and the other one simply goes into react/listen mode and stops trying to issue commands to the other one.

JamesMBallard (commented):
This check may only work on the initial interaction, but as the viewpoint keeps changing for an inertia pan in 3D (for example), both of them are now "navigating" so unclear if this works for all cases.

jared-2016:
> Ideally whichever one you interacted with last takes over and the other one simply goes into react/listen mode and stops trying to issue commands to the other one.

This is how it behaves with the latest changes. Inertia pan will maintain control until it's finished as well as zooming in an out. I think the evaluation order of the logical AND's works in our favor here.

If we do any interaction on the sceneview, we want to make sure there isn't an existing navigating on the mapview otherwise we won't try and upadate. If the mapview is not navigating and the sceneview is navigating, then we proceed with updating the mapview viewpoint.

Same goes for the inverse, we try to interact with the mapview. If the sceneview is nagivating(e.g. prior interaction on the sceneview) we don't update the viewpoint. If there is no navigating on the sceneview but the map is navigating, then we update the sceneview viewpoint.

I'll add a gif/video to the issue, the interactions seem natural to me. Scene has delayed navigation for some interactions which we would want to not be interrupted. MapQuickView doesn't really have anything like that besides maybe a zoom but those are quick so you don't really run into that same situation in the opposite direction.

CppSamples/Scenes/SyncMapViewSceneView/SyncMapViewSceneView.cpp
```diff
@@ -67,11 +67,19 @@ void SyncMapViewSceneView::setSceneView(SceneQuickView* sceneView)
...
+    {
       m_mapView->setViewpointAsync(m_sceneView->currentViewpoint(ViewpointType::CenterAndScale), 0);
+    }
+  });
+  connect(m_sceneView, &SceneQuickView::mousePressed, this, [this]
```
jared-2016, line 77: This ended up being the simplest solution to the interrupt problem. The viewpointChanged signal will only honor the one that is actively navigating as there will only be one. Yet if the other view is clicked we will interrupt that process allowing for either geoview to be interacted with.

I do have a branch with a flag and only setting up one signal handler etc. but it's an over engineered approach to me.

JamesMBallard: One last thought here. What about mouse wheel? That can still get around this `mousePressed` workflow I would think.

JamesMBallard (approved):
I don't want to overengineer this sample, so I think what you have is good. One last question, but I will defer to you on whether you think it needs further consideration or not.

CppSamples/Scenes/SyncMapViewSceneView/SyncMapViewSceneView.cpp
jared-2016: Ah good point. I naturally click the one i want to interact with first so it didn't cross my mind. Fix incoming.

aji13187 (approved):

PR #1920, opened 2025-11-18
Fix QML warnings

har13205:
# Description

Fix QML warnings

## Type of change

- [x] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

## Checklist

- [ ] Self-review of changes

### DIFF
--- a/CppSamples/EditData/SnapGeometryEdits/SnapGeometryEdits.qml
+++ b/CppSamples/EditData/SnapGeometryEdits/SnapGeometryEdits.qml
@@ -387,7 +387,11 @@ Item {
                             Layout.alignment: Qt.AlignRight
                             Layout.fillWidth: true
                             checked: isEnabled
-                            onCheckedChanged: isEnabled = checked;
+                            onCheckedChanged: {
+                                if (isEnabled !== checked) {
+                                    isEnabled = checked
+                                }
+                            }
                         }
                     }
                 }
--- a/CppSamples/Layers/ExportTiles/ExportTiles.qml
+++ b/CppSamples/Layers/ExportTiles/ExportTiles.qml
@@ -54,7 +54,6 @@ ExportTilesSample {
                 height: 24
                 color: palette.text
             }
-            visible: mapLoaded
 
             onClicked: {
                 // call the C++ invokable function to export tile cache from the input screen coordinates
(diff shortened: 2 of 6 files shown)


CppSamples/EditData/SnapGeometryEdits/SnapGeometryEdits.qml
```diff
@@ -387,7 +387,11 @@ Item {
                             Layout.alignment: Qt.AlignRight
                             Layout.fillWidth: true
                             checked: isEnabled
-                            onCheckedChanged: isEnabled = checked;
+                            onCheckedChanged: {
+                                if (isEnabled !== checked) {
+                                    isEnabled = checked
+                                }
```
JamesMBallard, line 393: This doesn't appear to address the underlying problem with the binding loop. Is it possible to make isEnabled and checked not depend on each other, or tweak the code some way other than this?

I don't have any suggestions, but it looks like `checked` binds to `isEnabled`, and `isEnabled` still kinda-sort changes with `checked` but no longer with a direct binding.

min14751 (approved):
LGTM!

CppSamples/EditData/SnapGeometryEdits/SnapGeometryEdits.qml
min14751: We do the same thing in a similar sample [here](https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/01be131366ee3d79dbe5a3a6694b8b67759818e7/CppSamples/EditData/SnapGeometryEditsWithRules/SnapGeometryEditsWithRules.qml#L295).

bri12415: > This doesn't appear to address the underlying problem with the binding loop

I'm not sure I follow. I think the guard statement is sufficient to prevent the binding loop isn't it? They are still tied to each other, but after the first setting of `isEnabled` to `checked`, their values will match and the guard statement will prevent the binding loop.

JamesMBallard: Yes this addresses the binding loop but there is conceptually still a circular dependency. `checked` depends on `isEnabled`, and `isEnabled` depends on `checked`. 

These changes address the warning, but is it possible to remove this circular dependency?

It's not mandatory, and it looks like we do the same thing elsewhere as well.

JamesMBallard (approved):
Not mandatory to resolve the circular dependency. The binding loop is resolved.

PR #1922, opened 2025-11-20
Update filter features sample

tanneryould:
# Description

The old OSM buildings layer is deprecated, and this updates the layer.
It also changes the order of operations in the sample such that the Esri buildings are removed before adding the detailed buildings.

## Type of change

- [x] Bug fix

**Platforms tested on**:

- [x] Windows

### DIFF
--- a/CppSamples/Scenes/FilterFeaturesInScene/FilterFeaturesInScene.cpp
+++ b/CppSamples/Scenes/FilterFeaturesInScene/FilterFeaturesInScene.cpp
@@ -147,15 +147,29 @@ void FilterFeaturesInScene::init()
 // Show the detailed buildings scene layer
 void FilterFeaturesInScene::showDetailedBuildings()
 {
+  if (!m_detailedBuildingsSceneLayer)
+  {
+    qWarning() << "Detailed San Fransisco Buildings layer not loaded";
+    return;
+  }
   m_detailedBuildingsSceneLayer->setVisible(true);
 }
 
 // Hide buildings within the detailed building extent so they don't clip
 void FilterFeaturesInScene::filterScene()
 {
+  if (!m_sceneView || !m_sceneView->graphicsOverlays()->first())
+  {
+    return;
+  }
 
   // Show the extent graphic to visualize the polygon filter
   m_sceneView->graphicsOverlays()->first()->graphics()->append(m_sanFranciscoExtentGraphic);
+
+  if (!m_3dBuildings || !m_sceneLayerPolygonFilter)
+  {
+    return;
+  }
 
   // Initially, the building layer does not have a polygon filter, set it
   if (!m_3dBuildings->polygonFilter())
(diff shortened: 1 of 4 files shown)


CppSamples/Scenes/FilterFeaturesInScene/FilterFeaturesInScene.cpp
```diff
@@ -55,24 +53,55 @@
...
-  m_scene = new Scene(basemap, this);
-
   // Construct and set the scene topography
-  Surface* surface = new Surface(this);
+  auto* surface = new Surface(this);
   surface->elevationSources()->append(new ArcGISTiledElevationSource(QUrl("https://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer"), this));
   m_scene->setBaseSurface(surface);
 
+  // Construct a basemap with the ArcGIS Navigation 3D Basemap then set it on the Scene
+  auto* basemap = new Basemap(QUrl("https://arcgisruntime.maps.arcgis.com/home/item.html?id=00a5f468dda941d7bf0b51c144aae3f0"), this);
```
ldanzinger: for samples we typically shy away from auto because it can make it a little less clear what type you are working with. Subjective, but that's what we've done, so probably best to stay consistent

koushik-hajra: I too believe in readability for our users. Using the actual `Type` to declare the objects keeps things more readable. In these cases the RHS gives you the type info right away but in other cases it won't. So, keeping it consistent is better.

CppSamples/Scenes/FilterFeaturesInScene/FilterFeaturesInScene.cpp
```diff
@@ -55,24 +53,55 @@
...
+    }
+    if (!m_3dBuildings)
+    {
+      qWarning() << "Buildings layer not found in base layers. Please check that your basemap"
+                 << "contains an ArcGIS Scene Layer with 'building' in the name";
+    }
+  });
+
   // Construct the detailed buildings scene layer
   m_detailedBuildingsSceneLayer = new ArcGISSceneLayer(QUrl("https://tiles.arcgis.com/tiles/z2tnIkrLQ2BRzr6P/arcgis/rest/services/SanFrancisco_Bldgs/SceneServer"), this);
```
koushik-hajra: I know it's been like that but I think It's better to change the name of this var. It is not a building scene layer and from reading it, it feels like it is that type. So, something like SanFrancisco3DOSceneLayer might be better.

CppSamples/Scenes/FilterFeaturesInScene/FilterFeaturesInScene.cpp
tanneryould: I remember that being the case, but the team's clang tidy linter was giving me this warning: `use auto when initializing with new to avoid duplicating the type name [modernize-use-auto]` and I wasn't sure if we had changed this.

ldanzinger: I see. I guess we will see if we need to re-evaluate when we run clang tidy against the entire codebase...

ldanzinger (approved):

koushik-hajra (approved):
Looks good. Thank you for updating the sample.

PR #1937, opened 2025-12-18
updating template to support local scene

ldanzinger:
Adding support for local scene view, based on the same changes made in the SDK templates. Also, removed a few relics from the past such as the feature layer input, which isn't used


aji13187 (approved):
LGTM

min14751 (approved):

PR #1955, opened 2026-03-04
Additional changes

tanneryould:
# Description

Suggests additional places where we may need to update the API names. I'm opening this PR against https://github.com/Esri/arcgis-maps-sdk-samples-qt/pull/1954 so we can evaluate if these cases are also needed.

I removed any mentions of the `Viewshed` API in the ViewshedLocation sample because we don't actually use that API in that sample.

Also `LineOfSightTargetVisibility` shouldn't be `LineOfSight::targetVisibility` because it's its own Class

```cpp
    connect(lineOfSight, &ExploratoryGeoElementLineOfSight::targetVisibilityChanged, this, [this](LineOfSightTargetVisibility targetVisibility)
    {
      // Select taxi whenever there is a clear line of sight from observer position.
      m_taxi->setSelected(targetVisibility == LineOfSightTargetVisibility::Visible);
    });
```


JamesMBallard (approved):
Thank you!

PR #1969, opened 2026-04-06
Add KML Photo Overlay to Sample & Refactor

tanneryould:
# Description

I set out to add a KML Photo Overlay node to the dataset and while doing so, ended up refactoring a significant amount of the code so that we can use more modern Qt practices. I cut out a significant portion of the Viewpoint handling code which was quite robust partially because the original .kmz document had quite a bit of missing data, and partially because I think we're able to handle KmlViewpoints much more gracefully now.

### DIFF
--- a/CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
+++ b/CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
@@ -41,14 +41,12 @@
 #include "SceneQuickView.h"
 #include "SpatialReference.h"
 #include "Surface.h"
+#include "Viewpoint.h"
 
 // Qt headers
 #include <QFuture>
 #include <QStandardPaths>
 #include <QUuid>
-
-// STL headers
-#include <algorithm>
 
 using namespace Esri::ArcGISRuntime;
 
@@ -80,7 +78,7 @@ ListKmlContents::ListKmlContents(QObject* parent /* = nullptr */) :
   // add the elevation source to the scene to display elevation
   m_scene->baseSurface()->elevationSources()->append(elevationSource);
 
-  const QString filepath = defaultDataPath() + "/ArcGIS/Runtime/Data/kml/esri_test_data.kmz";
+  const QString filepath = defaultDataPath() + "/ArcGIS/Runtime/Data/kml/esri_kml_sample_data.kmz";
 
   // create new KML layer
   m_kmlDataset = new KmlDataset(filepath, this);
@@ -96,12 +94,8 @@ ListKmlContents::ListKmlContents(QObject* parent /* = nullptr */) :
     }
 
     // recursively build tree to display KML contents
-    const auto nodes = m_kmlDataset->rootNodes();
-    for (KmlNode* node : nodes)
-    {
-      m_kmlNodesList << node;
-      buildTree(node);
-    }
+    const QList<KmlNode*> nodes = m_kmlDataset->rootNodes();
+    buildTree(nodes);
 
     // if at top node, then display children
     if (!m_kmlNodesList.isEmpty() && m_kmlNodesList[0]->parentNode() == nullptr)
@@ -160,10 +154,7 @@ QStringList ListKmlContents::buildPathLabel(KmlNode* node) const
   {
     return buildPathLabel(node->parentNode()) << node->name();
   }
-  else
-  {
-    return QStringList{};
-  }
+  return QStringList{};
 }
 
 void ListKmlContents::displayPreviousLevel()
@@ -179,18 +170,15 @@ void ListKmlContents::displayPreviousLevel()
   if (grandparentNode != nullptr)
   {
     m_currentNode = grandparentNode;
-    emit currentNodeChanged();
-    emit labelTextChanged();
-    emit levelNodeNamesChanged();
   }
   // if grandparent node is nullptr, then at top of tree
   else
   {
     m_currentNode = parentNode;
-    emit currentNodeChanged();
-    emit labelTextChanged();
-    emit levelNodeNamesChanged();
   }
+  emit currentNodeChanged();
+  emit labelTextChanged();
+  emit levelNodeNamesChanged();
 }
 
 // display selected node on sceneview and show its children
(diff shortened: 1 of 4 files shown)


CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -510,26 +364,20 @@ QStringList ListKmlContents::levelNodeNames()
...
     }
-  }
-
-  return hasNoGrandchildren;
+    else if (KmlNetworkLink* networkLink = dynamic_cast<KmlNetworkLink*>(node))
+    {
+      return !networkLink->childNodes().isEmpty();
+    }
+    return false;
+  });
```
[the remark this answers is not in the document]
tanneryould: Good catch, I changed `hasNoGrandchildren` to `hasGrandchildren` at the last minute, because every check of the method was calling `!hasNoGrandchildren` which was a confusing double negative to me.

CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -214,153 +202,108 @@ void ListKmlContents::processSelectedNode(const QString& nodeName)
...
-                                                                kmlViewpoint.pitch(), kmlViewpoint.roll()));
-        return;
+        return Camera(viewpoint.location(), viewpoint.range(), viewpoint.heading(), viewpoint.pitch(), viewpoint.roll());
+        break;
       case KmlViewpointType::Camera:
-        m_viewpoint =
-          Viewpoint(kmlViewpoint.location(), Camera(kmlViewpoint.location(), kmlViewpoint.heading(), kmlViewpoint.pitch(), kmlViewpoint.roll()));
-        return;
+        return Camera(viewpoint.location(), viewpoint.heading(), viewpoint.pitch(), viewpoint.roll());
+        break;
```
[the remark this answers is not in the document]
tanneryould: This was another last minute change that I didn't fully cleanup.

min14751 (approved):

CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -213,154 +205,115 @@ void ListKmlContents::processSelectedNode(const QString& nodeName)
...
-          m_sceneView->setViewpointAsync(m_viewpoint);
+          m_currentNode = container->childNodesListModel()->at(0);
+          emit currentNodeChanged();
         }
       }
 
-      // reset m_lastLevel before levelNodeNames() is called
-      emit levelNodeNamesChanged();
+      // If node has viewpoint, set view to that viewpoint. Otherwise, create a viewpoint to encompass node's geometry
+      KmlViewpoint nodeViewpoint = node->viewpoint();
```
JamesMBallard, line 221:
```suggestion
      const KmlViewpoint nodeViewpoint = node->viewpoint();
```

CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -213,154 +205,115 @@ void ListKmlContents::processSelectedNode(const QString& nodeName)
...
+        {
+          if (nodeExtent.width() == 0 && nodeExtent.height() == 0)
+          {
+            nodeViewpoint = KmlViewpoint::createLookAtViewpoint(nodeExtent.center(), 0, 45, 1000, KmlAltitudeMode::RelativeToGround);
+          }
+          else
+          {
+            nodeViewpoint = KmlViewpoint::createWithViewpoint(Viewpoint(nodeExtent));
+          }
+        }
```
JamesMBallard, line 237: Is this something that can happen, or is it just being defensive against possible problems?

If I read it right, this means we could have a valid extent from a node, but that extent can have 0 width and height. That can happen?

CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -213,154 +205,115 @@ void ListKmlContents::processSelectedNode(const QString& nodeName)
...
-  if (!kmlViewpoint.isEmpty())
+  if (viewpoint.isEmpty())
   {
-    // altitude adjustment is needed for all but Absolute altitude mode
-    m_needsAltitudeFixed = (kmlViewpoint.altitudeMode() != KmlAltitudeMode::Absolute);
+    return;
+  }
 
-    switch (kmlViewpoint.type())
+  const Camera camera = [viewpoint]() -> Camera
```
JamesMBallard:
```suggestion
  const Camera camera = [&viewpoint]() -> Camera
```
I don't think you need a copy.

CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -213,154 +205,115 @@ void ListKmlContents::processSelectedNode(const QString& nodeName)
...
-    {
-      onLocationToElevationCompleted_(elevation);
-    });
+    m_sceneView->setViewpointCameraAsync(camera, 1);
   }
 }
 
 // recursively build list of all KML nodes
-void ListKmlContents::buildTree(KmlNode* parentNode)
+void ListKmlContents::buildTree(QList<KmlNode*> rootNodes)
```
JamesMBallard: Is there any advantage of taking a copy here? Does const-ref have some disadvantage?

CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
tanneryould: Yes, this can happen if the node is a Placemark point, in which case it'll have an extent but xMin == xMax and yMin == yMax

CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
tanneryould: Oh, const ref would work here too, good catch

CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
JamesMBallard: That sounds odd, but ok. Thanks for the explanation.

CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
tanneryould: nodeViewpoint is overwritten in the following code block, but I can set this from a lambda to allow it to be const

JamesMBallard: No, it's fine like this. It might overcomplicate it just for the sake of making the local var const. Not worth it.

CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
tanneryould: You're not wrong, KML is indeed odd at times 😅

CppSamples/Layers/ListKmlContents/ListKmlContents.cpp
```diff
@@ -213,154 +205,115 @@ void ListKmlContents::processSelectedNode(const QString& nodeName)
...
+      KmlViewpoint nodeViewpoint = node->viewpoint();
 
-      // if displaying end-nodes, change m_currentNode to first end-node for correct behavior of back button
-      if (noGrandchildren(m_currentNode))
-      {
-        m_currentNode = static_cast<KmlContainer*>(m_currentNode)->childNodesListModel()->at(0);
-        emit currentNodeChanged();
-      }
+      if (nodeViewpoint.isEmpty())
+        {
```
JamesMBallard: Looks like some minor misalignment was introduced.

tanneryould: You were five seconds faster than me, thanks for catching this

JamesMBallard (approved):

PR #1985, opened 2026-04-22
Add api key option

har13205:
# Description

Adding an option to update API key from UI for testing purpose. Added an environmental variable `HIDE_APIKEY_OPTION`  to disable this option for final build.

## Type of change

- [ ] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [ ] iOS

## Checklist

- [ ] Self-review of changes

### DIFF
--- a/SampleViewer/SampleManager.cpp
+++ b/SampleViewer/SampleManager.cpp
@@ -522,7 +522,7 @@ bool SampleManager::hideApiKeyOption() const
 
 void SampleManager::setUserApiKey(const QString& key)
 {
-  const QString enteredApiKey = key;
+  const QString& enteredApiKey = key;
   if (userApiKey == enteredApiKey)
   {
     return;
--- a/SampleViewer/SampleManager.cpp
+++ b/SampleViewer/SampleManager.cpp
@@ -89,7 +89,6 @@ static QString homePath();
 namespace
 {
   QString apiKey = ""; // Provide your API key here
-  QString userApiKey = "";
 } // namespace
 
 SampleManager::SampleManager(QObject* parent) :
@@ -482,23 +481,15 @@ void SampleManager::setCurrentCategory(SampleCategory* category)
 
 void SampleManager::setApiKey(bool isSupportsApiKey)
 {
-  QString sampleApiKey;
+  QString sampleApiKey = apiKey;
 
-  if (!userApiKey.isEmpty())
-  {
-    sampleApiKey = userApiKey;
-  }
-  else
+  if (sampleApiKey.isEmpty())
   {
 #ifdef SAMPLE_VIEWER_API_KEY
     // If the API key identifier is defined in the respective .pro file it will be used here
     // Otherwise use the API key provided by the user at the top of this file
     sampleApiKey = QUOTE(SAMPLE_VIEWER_API_KEY);
 #endif
-    if (sampleApiKey.isEmpty())
-    {
-      sampleApiKey = apiKey;
-    }
   }
 
   if (isSupportsApiKey && sampleApiKey.isEmpty())
@@ -511,10 +502,10 @@ void SampleManager::setApiKey(bool isSupportsApiKey)
   ArcGISRuntimeEnvironment::setApiKey(keyToApply);
 }
 
-bool SampleManager::hideApiKeyOption() const
+bool SampleManager::showApiKeyOption() const
 {
-#ifdef HIDE_APIKEY_OPTION
-  return QString(QUOTE(HIDE_APIKEY_OPTION)) == "true";
+#ifdef SHOW_APIKEY_OPTION
+  return QString(QUOTE(SHOW_APIKEY_OPTION)) == "true";
 #else
   return false;
 #endif
@@ -523,12 +514,12 @@ bool SampleManager::hideApiKeyOption() const
 void SampleManager::setUserApiKey(const QString& key)
 {
   const QString& enteredApiKey = key;
-  if (userApiKey == enteredApiKey)
+  if (apiKey == enteredApiKey)
   {
     return;
   }
 
-  userApiKey = enteredApiKey;
+  apiKey = enteredApiKey;
 
   if (currentSampleSupportsApiKey())
   {
--- a/SampleViewer/SampleManager.cpp
+++ b/SampleViewer/SampleManager.cpp
@@ -35,6 +35,7 @@
 #include <QJsonObject>
 #include <QNetworkProxy>
 #include <QQmlEngine>
+#include <QRegularExpression>
 #include <QSet>
 #include <QStandardPaths>
 #include <QStringList>
@@ -89,7 +90,8 @@ static QString homePath();
 namespace
 {
   QString apiKey = ""; // Provide your API key here
-}
+  QString userApiKey = "";
+} // namespace
 
 SampleManager::SampleManager(QObject* parent) :
   QObject(parent),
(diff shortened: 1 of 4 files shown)


SampleViewer/SampleManager.cpp
```diff
@@ -89,7 +89,8 @@ static QString homePath();
 namespace
 {
   QString apiKey = ""; // Provide your API key here
-}
+  QString userApiKey = "";
+} // namespace
```
JamesMBallard, line 92: Do we need the concept of two API keys? One set during the build (potentially) and one set during runtime for testing? To me, I would think there can still be only a single key with the option for changing it via your new UI.

SampleViewer/SampleManager.cpp
```diff
@@ -472,19 +482,62 @@ void SampleManager::setCurrentCategory(SampleCategory* category)
...
+  ArcGISRuntimeEnvironment::setApiKey(keyToApply);
+}
+
+bool SampleManager::hideApiKeyOption() const
+{
+#ifdef HIDE_APIKEY_OPTION
+  return QString(QUOTE(HIDE_APIKEY_OPTION)) == "true";
+#else
+  return false;
+#endif
```
JamesMBallard: I would reverse this logic. Since it is primarily for debugging purposes, we should need to set something special to enable it, rather than needing to do something to disable it.

If users build this app, they'll see it.

SampleViewer/SampleViewer.pro
```diff
@@ -28,6 +28,7 @@ COMMONVIEWER = $$PWD
 PCH_HEADER = $$COMMONVIEWER/pch.hpp
 include($$PWD/../ArcGISRuntimeVersion.pri)
 DEFINES += ArcGIS_Runtime_Version=$$ARCGIS_RUNTIME_VERSION
+DEFINES += HIDE_APIKEY_OPTION=$$(HIDE_APIKEY_OPTION)
```
JamesMBallard: ditto I think this should be `DEFINES += SHOW_APIKEY_OPTION=$$(SHOW_APIKEY_OPTION)`

SampleViewer/SampleManager.cpp
har13205: Changed to single key.

SampleViewer/SampleManager.cpp
har13205: updated

SampleViewer/SampleViewer.pro
har13205: updated

SampleViewer/SampleManager.cpp
```diff
@@ -472,19 +481,54 @@ void SampleManager::setCurrentCategory(SampleCategory* category)
...
+#ifdef SHOW_APIKEY_OPTION
+  return QString(QUOTE(SHOW_APIKEY_OPTION)) == "true";
+#else
+  return false;
+#endif
+}
+
+void SampleManager::setUserApiKey(const QString& key)
+{
+  const QString& enteredApiKey = key;
```
JamesMBallard: Is this needed? key is already a const ref.

SampleViewer/SampleManager.cpp
```diff
@@ -472,19 +481,54 @@ void SampleManager::setCurrentCategory(SampleCategory* category)
...
+bool SampleManager::showApiKeyOption() const
+{
+#ifdef SHOW_APIKEY_OPTION
+  return QString(QUOTE(SHOW_APIKEY_OPTION)) == "true";
+#else
+  return false;
+#endif
+}
+
+void SampleManager::setUserApiKey(const QString& key)
```
JamesMBallard: Does this method need the distinction of a "user" api key? I think it can just refer to it as an Api key, and where it came from isn't important.

SampleViewer/SampleViewer.pro
```diff
@@ -28,6 +28,7 @@ COMMONVIEWER = $$PWD
 PCH_HEADER = $$COMMONVIEWER/pch.hpp
 include($$PWD/../ArcGISRuntimeVersion.pri)
 DEFINES += ArcGIS_Runtime_Version=$$ARCGIS_RUNTIME_VERSION
+DEFINES += SHOW_APIKEY_OPTION=$$(SHOW_APIKEY_OPTION)
```
JamesMBallard, line 31: You'll need to work with dev ops to setup a new process to ensure this new env var is part of the daily builds, and disabled at the end of the release.

On the other hand, @ldanzinger should we consider leaving it on all the time? There have been times where old API keys get invalidated and we can no longer use older builds that have no way to update the API key aside from rebuilding them. Something to consider, but not mandatory.

SampleViewer/SampleManager.h
```diff
@@ -104,6 +105,8 @@ class SampleManager : public QObject
   Q_INVOKABLE void setSourceCodeIndex(int i);
   Q_INVOKABLE void setupProxy(const QString& hostName, quint16 port, const QString& user, const QString& pw);
   Q_INVOKABLE void setApiKey(bool isSupportsApiKey = true);
+  Q_INVOKABLE bool showApiKeyOption() const;
+  Q_INVOKABLE void setUserApiKey(const QString& key);
```
JamesMBallard: This is going slightly beyond the scope of your changes, but I think you should rename the existing `setApiKey` method to something else. That's not what it actually does, and you need that function name now for that purpose.

Here is an idea
```c++
Q_INVOKABLE void setCurrentApiKeyToMapsSDK();
Q_INVOKABLE void unsetApiKeyFromMapsSDK();
Q_INVOKABLE void setApiKey(const QString& key);
```

The methods should convey clear intent. Hopefully those examples are clear and anyone reading the code would recognize what each one does. There are surely even better names.

SampleViewer/SampleManager.cpp
har13205: Fixed it.

SampleViewer/SampleManager.cpp
har13205: Agreed. I updated the name to `setApiKey`.

SampleViewer/SampleViewer.pro
har13205: yes, I had a word. I will create an issue with the details once I have this approval.

SampleViewer/SampleManager.h
har13205: Great suggestion. I refactored the API key methods.

SampleViewer/SampleManager.cpp
```diff
@@ -434,22 +434,31 @@ void SampleManager::setCurrentSample(Sample* sample)
...
+{
+  // NOTE - currently we know we cannot set an API Key for the
+  // following samples since they use named user login instead.
+  if (!m_currentSample)
+  {
+    return true;
+  }
+
+  const QString sampleName = m_currentSample->name().toString();
+  return sampleName != "Create and save map" && sampleName != "Add items to portal" && sampleName != "Search for web map by keyword";
```
jared-2016, line 459: this is a little hard for me to reason about. If I am the only one, feel free to ignore but i would rethink this for better readability.

jared-2016 (commented):
one minor thing. looks good

SampleViewer/SampleManager.cpp
```diff
@@ -470,23 +479,62 @@ void SampleManager::setCurrentCategory(SampleCategory* category)
...
+#ifdef SHOW_APIKEY_OPTION
+  return QString(QUOTE(SHOW_APIKEY_OPTION)) == "true";
+#else
+  return false;
+#endif
+}
+
+void SampleManager::setApiKey(const QString& key)
+{
+  const QString enteredApiKey = key;
```
JamesMBallard: My previous comment was not about the const-ref-ness of this, it was why do you need this local variable at all? You have `key` as the argument passed in. What is the purpose of `enteredApiKey`? You can simply rename the argument if you prefer that name.

har13205: Updated.

SampleViewer/SampleManager.cpp
har13205: The other approach I think of is may be we can use a `QSet<QString>` of sample names and set if the current sample is in it.

JamesMBallard: I do think that would be easier to follow, but for now it's the same set of three it was before, and it's only coming up now because the logic was moved around so it appears in the diff. 

IMO we can change it, but I think for now we can wait until the list grows a bit more and then move to a QSet.

JamesMBallard (approved):

jared-2016 (approved):

PR #1998, opened 2026-07-17
New sample: Geometry Editor interaction preview

JamesMBallard:
# Description

I've extensively tested on macOS and iOS, and ensured that the icons look good across the board and the functionality is correct.

## Type of change

- [ ] Bug fix
- [ ] Other enhancement

**Platforms tested on**:

- [ ] Windows
- [ ] Android
- [ ] Linux
- [x] macOS
- [x] iOS


CppSamples/EditData/DisplayGeometryEditorInformationDuringInteraction/DisplayGeometryEditorInformationDuringInteraction.qml
```diff
@@ -0,0 +1,163 @@
...
+                    ToolTip.visible: hovered
+                    ToolTip.text: text
+
+                    onClicked: sampleModel.undo()
+                }
+
+                AccentButton {
+                    text: qsTr("Redo")
+                    enabled: sampleModel.canRedo
+                    icon.source: "qrc:/Samples/EditData/DisplayGeometryEditorInformationDuringInteraction/redo-24.svg"
```
JamesMBallard, line 138: All icons are standard calcite.

https://esri.github.io/calcite-ui-icons/

CppSamples/EditData/DisplayGeometryEditorInformationDuringInteraction/DisplayGeometryEditorInformationDuringInteraction.qml
```diff
@@ -0,0 +1,163 @@
...
+                Label {
+                    text: sampleModel.interactionDescription
+                    wrapMode: Text.WordWrap
+                }
+
+                Label {
+                    text: sampleModel.interactionValue
+                    // ensure the size of the text displayed doesn't
+                    // judder as the numbers rapidly change on interacting
+                    font.features: ({ "tnum": 1 })
```
JamesMBallard, line 163: This is something I learned thanks to copilot. I didn't like that the text would judder during interaction, like this... You can see that the width of the numbers aren't uniform, so the text judders rapidly.

https://github.com/user-attachments/assets/94d536e3-a04c-4b6a-bd28-6bcb56d1eb6f

But with this feature, we're asking for tabular numbers that are equal widths:

https://github.com/user-attachments/assets/9d059e6e-928e-47c5-bbd2-a06dc3c8594b

CppSamples/EditData/DisplayGeometryEditorInformationDuringInteraction/DisplayGeometryEditorInformationDuringInteraction.qml
```diff
@@ -0,0 +1,163 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// [Legal]
+
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Controls.Basic as Basic
```
JamesMBallard, line 19: This is to prevent warnings when running the sample standalone outside the viewer, which doesn't have our Calcite style applied.

CppSamples/EditData/DisplayGeometryEditorInformationDuringInteraction/DisplayGeometryEditorInformationDuringInteraction.qml
```diff
@@ -0,0 +1,163 @@
...
+        mapView: mapView
+    }
+
+    component AccentButton: Basic.Button {
+        id: accentButton
+
+        display: AbstractButton.IconOnly
+        icon.width: 32
+        icon.height: 32
+        icon.color: "#ffffff"
```
JamesMBallard, line 44: I found that in some cases the icon color could be black, and displayed over the calcite purple it was hard to see. So force white.

CppSamples/EditData/DisplayGeometryEditorInformationDuringInteraction/DisplayGeometryEditorInformationDuringInteraction.cpp
```diff
@@ -0,0 +1,341 @@
...
+void DisplayGeometryEditorInformationDuringInteraction::onInteractionPreviewChanged(GeometryEditorInteractionPreview* interactionPreview)
+{
+  if (!interactionPreview)
+  {
+    clearInteractionInformation();
+    return;
+  }
+
+  // Clean these up, we don't need them after we've processed the update.
+  // They won't leak, but they're owned by the GeometryEditor itself.
```
GuillaumeBelzPro: > but they're **owned** by the GeometryEditor itself.

Maybe "owned" is confusing. When I read that, I thought maybe this can create a dangling pointer in GeometryEditor. But it's just a parent-child relationship. 

```
// The preview is parented to GeometryEditor, so it would be cleaned up automatically.
// We can delete it earlier after this callback because we do not retain it beyond this update.
```
?

JamesMBallard: I like your comment better, I will change to that. Thank you. I agree "own" may be confusing here.

har13205 (approved):

GuillaumeBelzPro (approved):

